PR #63, opened 2023-09-01
Fix some xwayland windows not reappearing when forcing wayland rendering

### LINKED ISSUE #38: Some xwayland apps, when forced to use wayland, won't reappear on tag change
The three I've currently found are Discord and VS Code using the `--ozone-platform-hint=wayland` flag, as well as Firefox with the `MOZ_ENABLE_WAYLAND=1` environment variable.

For example, opening Wayland Firefox on a tag, switching to another, then switching back *once* requires a click on the empty space to get Firefox to reappear. If the tag Firefox is on is switched to twice or more in succession, the click won't make it reappear at all. Additionally, causing a resize by eg. opening a new window will cause the window to reappear.

Ottatop:
Fixes #38

### DIFF
--- a/src/backend/udev.rs
+++ b/src/backend/udev.rs
@@ -853,38 +853,6 @@ impl State {
             device_id: node,
         });
 
-        // Run any connected callbacks
-        {
-            let clone = output.clone();
-            self.loop_handle.insert_idle(|data| {
-                crate::state::schedule(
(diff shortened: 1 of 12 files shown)


Ottatop:
Whaddya know it was my flawed logic that was the issue

PR #100, opened 2023-10-03
Clean up udev file

Ottatop:
It's 1500 lines of stuff I copied from Anvil and there are some things that don't need to be there.


Ottatop:
Well it's still a 1500 line file but at least I understand it a bit more :+1:

PR #110, opened 2023-12-10
Attempt a codebase restructure

Ottatop:
Gonna try to make an actual, meaningful attempt to refactor the codebase before fixing stuff or adding new features as it feels like stuff's getting out of hand


Ottatop:
As always every time I try to do one of these restructures I end up moving stuff around and not actually fixing architecture things like how resizing is currently implemented.

But I'm bored of this so :shipit:

Issue #147, opened 2024-01-25
Poor config performance


Ottatop:
Having a bunch of windows open and then calling `window.close` from a config generates a lot of gRPC calls. This is because `window.get_focused` searches all windows to find the focused one, calling `window.props` on each.

It takes the compositor way too long to process all of these calls.

Ottatop:
Alright, I found the issue:

# The Issue

Most API calls are blocking to give users synchronous reasoning of the program flow.

Let's look at `window.get_focused` as an example. It's implemented to call `window.get_all` and *synchronously* call `window.props` with each of the returned window_ids. This means that *each call* to `window.props` makes the call, then the compositor has to process it and return results, then the config can move on calling the next `window.props`. This is the first part of the problem.

The second part of the problem is that because of how the compositor's event loop works, *all* API calls take at least one cycle of the loop to actually get run and return results. Therein lies the second part of the problem.

Combining these two issues means that calls to `window.get_focused` take *as many event loop cycles as there are windows*.
With a few windows this isn't much of a problem; computers are fast enough. However...

Spawning lots and lots of windows will slow the event loop down. In debug mode this is *very* noticeable; in release mode less so. Unfortunately, this means that the time it takes to run `window.get_focused` scales not just off of the number of windows, but also the length of time each event loop cycle takes to run.

This was actually also a problem with the old MessagePack API that I never caught.

Now let's discuss proposed solutions.

# Solutions
The easiest thing to do would be to find a way to decouple dealing with gRPC calls from the event loop. Unfortunately, that seems to be out of the question as the event loop is the only sane way to retrieve the `&mut State` needed to fulfill gRPC requests.

One solution I'm currently experimenting with is batch-sending API calls to the compositor and awaiting them client-side. This means the client doesn't have to synchronously call `window.props` while still preserving user-facing synchronicity. This is easy to implement internally, but if the user wants to use this batch-sending strategy there would have to be some auxiliary function, which increases config complexity, something I'm trying to avoid. This solution also involves changing all API calls that expect `google.protobuf.Empty` to also asynchronously call. I'm not sure how this affects user-facing synchronicity, though--still have to test that.

Ottatop:
Superseded by #163

PR #158, opened 2024-02-08
Add signal system

Ottatop:
Should work similarly to Awesome's signal system. Hopefully.

Stuff to do before continuing:
- [x] Update dependencies
- [x] Change Lua API to differentiate between modules and objects
    - This was how it was originally, but I read some Luarocks style guides and they said global state was bad. This led me to have everything in the rewritten API be an object, which I am finding looks noisy and is a little confusing on what uses `.` vs `:`. So screw it we're bringing back global state. This also just makes writing the API easier.


Ottatop:
Currently debating whether or not to make that Clippy review workflow run on trigger because wow that's annoying

Ottatop:
Stalled on #147

Ottatop:
Superseded by #164

PR #200, opened 2024-04-11
Replace shell.nix with a updated flake.nix

LogicalOverflow:
Replaces the `shell.nix` with a flake, that provides a devShell with the libraries and tools need to build pinnacle. It also sets up the `LD_LIBRARY_PATH` so pinnacle can start.

It does not include a nix package for pinnacle (yet), as the build process isn't quite trivial (with it generating configs and doing lua things I don't know anything about). I might tackle that at a later time.


flake.nix
```diff
@@ -0,0 +1,59 @@
+{
+  description = " A WIP Smithay-based Wayland compositor, inspired by AwesomeWM and configured in Lua or Rust Topics";
```
Ottatop:
```suggestion
  description = " A WIP Smithay-based Wayland compositor, inspired by AwesomeWM and configured in Lua or Rust";
```

Copy paste error, I presume

Ottatop:
Oh also rebase/merge with main and we should be good

flake.nix
LogicalOverflow: Indeed; fixed

LogicalOverflow:
Just rebased the PR on main

Ottatop (approved):

Ottatop:
Awesome, thank you!

PR #204, opened 2024-04-16
Add wlcs tests

### LINKED ISSUE #128: Add end-to-end testing with WLCS
WLCS, the Wayland Conformance Test Suite, is used to test Wayland compositors apparently. So uh let's get that happening sometime this decade.

LogicalOverflow:
WIP, but running wlcs is already possible and it actually does something (marginally) useful. It's mostly waiting for test to fail after some timeout, but some tests do pass :D

It's based off of smithay's `wlcs_anvil`, the wlcs integration module for anvil.

I currently use this to run the tests (`wlcs` is included in the flake):
```
cargo build -p wlcs_pinnacle_config -p wlcs_pinnacle
PINNACLE_WLCS_CONFIG_PATH="./wlcs_pinnacle/config" RUST_BACKTRACE=1 RUST_LOG=warn \
    wlcs target/debug/libwlcs_pinnacle.so
```

I am trying to keep my changes limited to the wlcs integration and keep changes outside of the `wlcs_pinnacle` folder small.

Will (hopefully, eventually) fix #128


LogicalOverflow:
This should be mostly done now. I'll try to write a summary of how this works and what I don't like about my implementation tomorrow

Ottatop:
Btw tests seem to be hanging because the name of the Dummy output was changed. In hindsight I probably should've extracted that into a const.

LogicalOverflow:
> Btw tests seem to be hanging because the model of the Dummy output was changed. Need to find a way to extract stuff like that and interpolate it into the Lua tests.

Ah, that explains that. I was mildly confused by those failures. (Hopefully) fixed that now.

LogicalOverflow:
Here a quick rundown of the changes: All it all, it was quite straight forward: Give wlcs a way to start and stop Pinnacle and a way to do some mouse and touch interactions.

In pinnacle itself I added a new feature flag `wlcs`, which adds a field to the `Dummy` backend in which I store the state I need to manage for wlcs. All the other changes (besides making two things `pub(crate)`) are limited to the `backend::wlcs`.

The `setup_wlcs_dummy` function is a duplication of `setup_dummy`, which does not start xwayland. I would consider this a temporary workaround until "running Pinnacle without xwayland" is supported.

The `start_wlcs_config` is used to start the embedded config used during the tests; it's a stripped down version of `start_config`. If the config-startup ever gets a re-worked, we can probably merge the two.

(The reason I want to run Pinnacle without xwayland for these test is, that Pinnacle gets started and stop a lot during the tests, and that leads to a lot of locking and unlocking of `DISPLAY` values (as well as setup/teardown of xwayland itself), which I kindof want to avoid.)

The `wlcs_pinnacle` package is the core of the PR; it builds into the integration module for wlcs. `lib.rs` mostly just forwards the functions, called by wlcs, to events I pass into the event loop and handled in `main_loop.rs`. The initial setup of Pinnacle is a little bit scuffed, because I am not starting xwayland.

For the mouse and touch interactions, implementing a trivial `InputBackend` (in `input_backend.rs`) allows me to just throw the events into `state.process_input_event`, for the most part, which is quite nice, since this way, it should behave just like real user interactions.

The most interesting case is probably the `PositionWindow` handling. Positioning seems to work, but their sizes are, at least sometimes, 0 by 0. After some trying I haven't found a way to get my hands on the correct window size. I think
fixing this will probably require some changes in pinnacle itself to keep track of a clients preferred window dimensions or something, so I left it like this for now.

And `config.rs` contains the embedded config used for the tests. It just floats all windows and provides a layout manager.

All in all, I'd say the code in `wlcs_pinnacle` is quite boring, which feels like a good thing for a test integration module :D

I am currently running the whole wlcs suite, which takes ~1.5h for me, locally. I'll link to a gist with the output here, once it's done.

And to run the thing, you can use something like
```
cargo build -p wlcs_pinnacle 
RUST_BACKTRACE=1 RUST_LOG=warn wlcs target/debug/libwlcs_pinnacle.so
```

LogicalOverflow:
Logs from a complete wlcs run are here: https://gist.github.com/LogicalOverflow/804f83c870da2a7667fb9f6d3d915043
There two types of panics in there; investigating these will probably be the next thing I look into after this is merged.

For reference: There currently are
 - 673 failed,
 - 130 passed, and
 - 357 skipped tests
 
 of 1160 test in total.

wlcs_pinnacle/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+
+[dependencies]
+smithay = { workspace = true }
+pinnacle = { path = "..", features = [ "wlcs" ] }
+pinnacle-api = { path = "../api/rust" }
+wayland-sys = { version = "0.31.1", features = ["client", "server"] }
+wlcs = "0.1"
+
+tracing = "0.1.40"
+tracing-subscriber = { version = "0.3.18", features = ["env-filter", "registry"] }
```
Ottatop: We could probably make these workspace dependencies

wlcs_pinnacle/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+
+tokio = { workspace = true, features = ["net", "rt"] }
+
+tempfile = "3.10.1"
```
Ottatop: Same with this

flake.nix
```diff
@@ -31,6 +31,17 @@
...
+            # wlcs
+            (writeScriptBin "wlcs" ''
+              #!/bin/sh
+              ${wlcs}/libexec/wlcs/wlcs "$@"
+            '')
+
+            # build time stuff
+            pkg-config
+            protobuf
+            luarocks
```
Ottatop:
```suggestion
```

pkg-config was moved to native build deps and the other two are below

Ottatop (changes requested):
Just a few changes

Ottatop:
First off you're a legend, thanks for taking the time to do this!

> The `setup_wlcs_dummy` function is a duplication of `setup_dummy`, which does not start xwayland. I would consider this a temporary workaround until "running Pinnacle without xwayland" is supported.

Yea I'm not a fan of the duplication either. I've got some ideas to implement "no-xwayland" but that's for a future commit.

> Positioning seems to work, but their sizes are, at least sometimes, 0 by 0.

Interesting, maybe those are part of the tests? If not I'll have to do some investigation.

> which takes ~1.5h for me

Yeesh, I wish there was a way to lower the test timeout but there doesn't seem to be short of patching wlcs.

> Let me know if you want me to squash this down to a single commit. The commits in this PR do feel a bit noisy

I think they're fine, I push a bunch of crap all the time

> There two types of panics in there

One looks to be the layout code panicking because the layout receiver is dropped, probably when wlcs takes down an instance of Pinnacle. Might be best to make that just print the error.

The other seems to be a smithay issue and possible bug, intriguing.

> - 673 failed,
> - 130 passed

Yikes lmao

For the most part this PR looks great! Working on being able to disable xwayland and possibly modularizing `start_config` to take in different embedded configs would probably be the next logical step in streamlining this.

Once you're done with the review I'll push a `compile_wlcs.sh` script that I stole from smithay to simplify getting wlcs and we should be good to merge!

wlcs_pinnacle/Cargo.toml
LogicalOverflow: Done

wlcs_pinnacle/Cargo.toml
LogicalOverflow: Done

flake.nix
LogicalOverflow: And done

LogicalOverflow:
> One looks to be the layout code panicking because the layout receiver is dropped, probably when wlcs takes down an instance of Pinnacle. Might be best to make that just print the error.



> >  Positioning seems to work, but their sizes are, at least sometimes, 0 by 0.
>
> Interesting, maybe those are part of the tests? If not I'll have to do some investigation.

The test I was looking at were the ClientSurfaceEventsTest ones (You can use the ` --test_filter='ClientSurfaceEventsTest.*'` flag to run just those), as I expected those to pass, and they try to create non-zero-size windows, so somewhere that info gets lost. (https://github.com/canonical/wlcs/blob/main/tests/test_surface_events.cpp the calls to `client.create_visible_surface`)

Ottatop (approved):

Ottatop:
> The test I was looking at were the ClientSurfaceEventsTest ones

Figured it out I *think*: Pinnacle only resizes floating windows if they are on an active tag, and the wlcs_config doesn't add a tag and set it active. Adding that causes 3/6 of the ClientSurfaceEventsTests to pass instead of none at all.

LogicalOverflow:
> Figured it out I think: Pinnacle only resizes floating windows if they are on an active tag, and the wlcs_config doesn't add a tag and set it active. Adding that causes 3/6 of the ClientSurfaceEventsTests to pass instead of none at all.

Ah, nice! I suspect there are probably a few more tests that have similarly easy resolutions. But now that we have the tests, anyone can just look for failed test that look like they should work, and then debug why they don't. I'll probably do that myself from time to time

PR #226, opened 2024-04-29
Restart xwayland if it exits

### LINKED ISSUE #186: Restart xwayland if it crashes
Currently if xwayland crashes or exits by someone killing the Xwayland process then xwayland is just dead. Ideally we'd attempt a restart, but that creates a problem: restarting the xwayland server will increment `DISPLAY` by 1. It should be easy to send a message to a config to get it to set the new `DISPLAY`, but any already open applications can't get the updated `DISPLAY`. Unsure of how to deal with that.

LogicalOverflow:
Needs some (manual) testing, but I am optimistic about that. I'll un-draft the PR once I tested it

Fixes #186


Ottatop:
Can you target this PR against the [`pointer_constraints`](https://github.com/pinnacle-comp/pinnacle/tree/pointer_constraints) branch? Smithay recently updated the process of starting xwayland and this branch contains that update (which probably should've been a separate commit, I know). I can hold off on pushing commits until this PR is merged.

LogicalOverflow:
Will do. I'll try to do that either today or tomorrow

LogicalOverflow:
Okay, with the Smithay changes, I don't think this is currently possible, as I can't see a way to get notified that Xwayland exited. I'll re-try implementing this, once there's a way to do that, see https://github.com/Smithay/smithay/issues/1410

PR #227, opened 2024-05-02
Add security-context-v1

LogicalOverflow:
I can run pinnacle with it and stuff still works, and the code looks reasonable, but not sure yet how to "properly" test it yet, as I don't yet know who uses this protocol or how I can verify it works as it should

And not a huge fan of the `is_client_restricted` being defined on `State`, but also haven't come up with a better place to put it either

### DIFF
--- a/src/handlers.rs
+++ b/src/handlers.rs
@@ -3,14 +3,14 @@
 mod xdg_shell;
 mod xwayland;
 
-use std::{mem, os::fd::OwnedFd, time::Duration};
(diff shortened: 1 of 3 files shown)


Ottatop (approved):

Ottatop:
Aight lgtm

> And not a huge fan of the `is_client_restricted` being defined on `State`, but also haven't come up with a better place to put it either

Probably doesn't matter whether it's an associated method, free function, or closure; at that point we're getting into the pedantics

Other than that I don't see anything wrong soooo :shipit:

PR #238, opened 2024-05-15
Implement session lock

Ottatop:
NOT blatantly plagiarized from Niri this time gj me

TODO:
- [x] Force keyboard focus to always be on a lock surface
- [x] Block keybinds from executing when locked
- [x] Prevent pointer focus to windows when locked/locking without a lock surface
- [x] Maybe draw the background in a different color while locking? ~~too lazy yet again~~ fine I won't be lazy this time
- [x] Test locked output hotplug


ildar:
This is great news. Thank you!
Is there an upstream (Smithay) bug related to the errors?

Ottatop:
> Is there an upstream (Smithay) bug related to the errors?

Possibly. I have a fix that will allow gtklock to work but I'll have to confirm if it's the right thing to do.

PR #240, opened 2024-05-21
Add layout transactions

### LINKED ISSUE #118: Flickering when closing windows, unfullscreening, and a bunch of other things
#109 removes the hack where I straight up froze the screen while waiting for windows to resize and stuff when tiling, but this introduces the issue where closing a window causes it to disappear before all the other windows resize in time, leading to slight flickering.

Ottatop:
Holy cow no more flickering, this is awesome

## Introducing layout transactions!

You know how there's always a frame or two where you can see the background pop in when opening and closing windows? That's because all the windows commit their buffers at different times leading to times when a window has shrunk or disappeared but another window hasn't mapped or resized yet. Those days are now behind us!

The idea: have layout applications be transactional--the new layout will only be shown once its windows have committed their buffers. Of course, now we need something to show on screen while they're doing that work. That's why we need to take a snapshot of the windows that are currently on screen for display. We render those snapshots until all windows have committed, then swap to rendering the windows. Easy!

But what if a window is slow or unresponsive? Simple: we add a timeout and end the transaction if it takes too long. You can actually see Sway doing this if you have a video playing while wleird-slow-ack-configure is running and you open a new window: everything pauses for a bit then resumes after the timeout.

TODO:
- [x] Make transactions work with non-1 output scales
- [x] Test with different output transforms
- [x] Test with wleird-slow-ack-configure
- [x] Extract all the duplicate code
- [x] "Fix" the assert in `on_event_loop_cycle_completion`
    - aka remove it 
- [x] Clean up handling of new windows
- Most likely stop tiled windows from rendering and getting pointer input on other outputs
    - It's either this or extend transactions compositor-wide instead of output-local
    - It also allows fancier layouts that want to map stuff offscreen without disrupting other outputs
    - Split off into #242
- [x] Make it work with fullscreen, floating, maximized toggling
- [x] Test/make it work when moving windows to other tags
- [x] README: add acknowledgments to Niri because I've stolen enough code lmao
- [x] Daily drive for ~~a day~~ an hour or two instead of just merging like you always do
- [x] Probably other stuff that I'll need to remember

Fixes #118 (finally)


Ottatop:
Blocking since I have no clue what I'm doing and if I push anything I've written here maintaining anything will be impossible

Need to do some documentation reading and brain embiggening first

PR #253, opened 2024-06-13
Preliminary snowcap integration (Rust)

Ottatop:
This PR adds preliminary integration for [Snowcap](https://github.com/pinnacle-comp/snowcap), the very, *very* WIP widget system for Pinnacle. Snowcap integration is gated by the `snowcap` feature, enabled by default.

Snowcap is currently, for the most part, a thin API wrapper around [Iced](https://iced.rs/). As such, the current Rust API closely mirrors their widget API.

The only currently implemented widgets are row, column, scrollable, container, and text. Using those, a goal for this PR is to implement an exit prompt and a keybind list a la Awesome.

The Lua side has not been implemented yet and will be the target of a future PR.

TODO:
- [x] Add a builtin exit prompt
- [x] Update the keybinding API to add descriptions
    - [x] Then add an API call to list all keybinds and descriptions
        - [x] Then add a builtin keybind list
- [x] Split the default Rust config into `snowcap` and non-`snowcap` variants for the feature gate
    - ~~Don't really like this tbh~~ found a decent way to do it yay
- [x] Document <sup>some of</sup> the API <sup>(i'm lazy)</sup>
- [x] Fix Rust config with Snowcap support eating up 100% of a CPU thread

### DIFF
--- a/.github/workflows/ci.pinnacle.yml
+++ b/.github/workflows/ci.pinnacle.yml
@@ -17,6 +17,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+        with:
+          submodules: recursive
       - name: Get Rust toolchain
         uses: dtolnay/rust-toolchain@stable
       - name: Cache stuff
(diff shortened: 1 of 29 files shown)


Ottatop:
Fixes #254

Issue #138, opened 2024-01-23
Improve libinput settings handling


Ottatop:
- [ ] Check what type a device is (trackpad, mouse, etc) before actually applying libinput settings
- [ ] Add the ability to target a specific device to set settings for in the API
- [ ] Add send events mode
- [ ] Apply default settings when possible

PR #271, opened 2024-12-31
Make APIs better

Ottatop:
~~Well my right middle finger's knuckle is still sore from no-lifeing Celeste so I decided to work on this thing instead~~ we are so back baby

This PR is intended to be the big final breaking change to the Lua and Rust APIs until 0.1 is released. As a result, this PR is going to be way too big and cover the following aspects:
- [x] Update the protobuf from proto2 to proto3
- [x] Better window rule API
- [x] Better layout API
    - Not a fan of how the client is expected to calculate its own rects for layouting; this means I have to duplicate all the logic over Lua and Rust, and it doesn't support resizing tiles
    - Something compositor-side is gonna be needed
    - Alternatively push this off to some future update
- [x] Better output management API, or simply remove the current one and direct you to something like [kanshi](https://git.sr.ht/~emersion/kanshi) until I can figure something out
- [x] Better process API
    - [x] Better spawning API for Rust that mirrors `std::process::Command`
    - [x] Something akin to `awful.spawn.once` so you can spawn things on startup, close them, and not have them respawn on config reload
- [x] Better libinput API
- [x] Better key- and mousebind API
- [x] Hide the snowcap API until it's actually somewhat stable
- [x] Probably a bunch of other things
- [x] And try to fix all the FIXMEs you left everywhere
- [x] And update the config gen cli
- ~~And remember to do that "let xwayland scale itself" thing~~ pushing off to later

~~Side note: I regret deciding to do a Rust API because wow is it not fun to try to design something remotely ergonomic haha~~ nvm this is kinda fun

Edit: hey past me it's present me here to tell you to:
- ~~Add docs for the move from metaconfig.toml to pinnacle.toml~~
- ~~Add docs for the new input API~~
- DOCS SOON I SWEAR
- [x] Oh and do this all again for the Lua side :+1:


Ottatop:
Closes #138

PR #274, opened 2025-01-29
Add window focused signal

anriha:
This PR adds a new signal that gets triggered whenever a window receives keyboard focus. I've tested the core functionality and it works as expected, though I haven't tested the Lua API yet (it's a bit tricky to set up on NixOS).

This is just the first of several signals I'm planning to add. My goal is to implement enough signals to replicate the dwl format for yambar integration, as I'd like to make pinnacle my daily driver.

Personal note: I'm really excited about this project! I've tried many Wayland compositors, and was actually planning to write my own because each existing one was missing something crucial. Then I found pinnacle, and it's exactly what I was looking for. Really happy to be able to contribute.

### DIFF
--- a/api/rust/src/signal.rs
+++ b/api/rust/src/signal.rs
@@ -218,6 +218,21 @@ signals! {
                 }
             },
         }
+        /// The window got keyboard focus.
+        ///
+        /// Callbacks receive the window gets focus.
+        WindowFocused = {
+            enum_name = Focused,
+            callback_type = SingleWindowFn,
+            client_request = window_focused,
+            on_response = |response, callbacks| {
+                let handle = WindowHandle { id: response.window_id };
+
+                for callback in callbacks {
+                    callback(&handle);
+                }
+            },
+        }
     }
     /// Signals relating to tag events.
     TagSignal => {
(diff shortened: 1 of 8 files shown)


api/rust/src/signal.rs
```diff
@@ -213,6 +213,21 @@ signals! {
...
                 let handle = WindowHandle { id: response.window_id };
 
+                for callback in callbacks {
+                    callback(&handle);
+                }
+            },
+        }
+        /// The window got keyboard focus.
+        ///
+        /// Callbacks receive the window gets focus.
```
Ottatop:
```suggestion
        /// Callbacks receive the newly focused window.
```

Ottatop (changes requested):
Oh hey I was thinking about this a few days ago

Implementation looks good! Did a sanity check and it does work on the Lua side. Just a minor comment nit.

I was *considering* whether or not we could extend this to something like `fn(Option<&WindowHandle>,  Option<&WindowHandle>)` and give people both the old and new focuses as well as notifying on focus loss, but I'm not sure how to handle the case where a window gets closed. In that case I don't really want to pass a dead window handle to the signal because it's not useful. We can roll with this until something gets figured out.

Ottatop (approved):

Ottatop:
Aight lgtm, thank you!

PR #278, opened 2025-02-04
Replace deprecated license identifier (`GPL-3.0+` → `GPL-3.0-or-later`)

PatchMixolydic:
According to [The Cargo Book](https://doc.rust-lang.org/cargo/reference/manifest.html#the-license-and-license-file-fields):
> [crates.io](https://crates.io/) interprets the license field as an [SPDX 2.3 license expression](https://spdx.github.io/spdx-spec/v2.3/SPDX-license-expressions/). The name must be a known license from the [SPDX license list 3.20](https://github.com/spdx/license-list-data/tree/v3.20). See the [SPDX site](https://spdx.org/) for more information.

According to the SPDX license list 3.20, [the identifier `GPL-3.0+` is deprecated](https://github.com/spdx/license-list-data/blob/v3.20/text/deprecated_GPL-3.0%2B.txt). This patch replaces it with the modern equivalent, `GPL-3.0-or-later`.


Ottatop (approved):
LGTM, thank you!

PR #279, opened 2025-02-07
Fixes source output layout not being updated when moving window to tag on another output

anriha:
This fixes an issue when trying to move window to tag on another output. Currently, only the layout of the target output gets updated. This updates both the source and target layout.

### DIFF
--- a/src/api/window.rs
+++ b/src/api/window.rs
@@ -232,6 +232,10 @@ pub fn move_to_tag(state: &mut State, window: &WindowElement, tag: &Tag) {
         state.tags = std::iter::once(tag.clone()).collect();
     });
 
+    if !is_window_mapped(window) {
+        return;
+    }
+
     if let Some(output) = source_output.as_ref() {
         state.pinnacle.begin_layout_transaction(output);
         state.pinnacle.request_layout(output);
@@ -239,21 +243,18 @@ pub fn move_to_tag(state: &mut State, window: &WindowElement, tag: &Tag) {
         state.schedule_render(output);
     }
 
-    if !is_window_mapped(window) {
-        return;
-    }
-
     let Some(target_output) = tag.output(&state.pinnacle) else {
+        state.pinnacle.update_xwayland_stacking_order();
         return;
     };
 
-    if let Some(source_output) = source_output {
-        if source_output != target_output {
-            state.pinnacle.begin_layout_transaction(&target_output);
-            state.pinnacle.request_layout(&target_output);
+    if source_output.as_ref() != Some(&target_output) && tag.active() {
+        state.capture_snapshots_on_output(&target_output, [window.clone()]);
 
-            state.schedule_render(&target_output);
-        }
+        state.pinnacle.begin_layout_transaction(&target_output);
+        state.pinnacle.request_layout(&target_output);
+
+        state.schedule_render(&target_output);
     }
 
     state.pinnacle.update_xwayland_stacking_order();
--- a/src/api/window.rs
+++ b/src/api/window.rs
@@ -222,9 +222,9 @@ pub fn set_decoration_mode(
 }
 
 pub fn move_to_tag(state: &mut State, window: &WindowElement, tag: &Tag) {
-    let output = window.output(&state.pinnacle);
+    let source_output = window.output(&state.pinnacle);
 
-    if let Some(output) = output.as_ref() {
+    if let Some(output) = source_output.as_ref() {
         state.capture_snapshots_on_output(output, [window.clone()]);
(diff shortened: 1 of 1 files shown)


src/api/window.rs
```diff
@@ -222,28 +222,39 @@ pub fn set_decoration_mode(
...
+    if let Some(output) = source_output.as_ref() {
+        state.pinnacle.begin_layout_transaction(output);
+        state.pinnacle.request_layout(output);
+
+        state.schedule_render(output);
+    }
+
     if !is_window_mapped(window) {
         return;
     }
```
Ottatop: Can you move the layout `if` block below the `is_window_mapped` one? If the window isn't mapped then we don't need to spend time updating the layout.

src/api/window.rs
```diff
@@ -222,28 +222,39 @@ pub fn set_decoration_mode(
...
-    state.pinnacle.request_layout(&output);
+    if let Some(source_output) = source_output {
+        if source_output != target_output {
+            state.pinnacle.begin_layout_transaction(&target_output);
+            state.pinnacle.request_layout(&target_output);
 
-    state.schedule_render(&output);
+            state.schedule_render(&target_output);
+        }
+    }
```
Ottatop: If the window didn't have any tags then this won't trigger a layout on the new output. We can remove the outer if statement and just check `if source_output != Some(target_output)`.

src/api/window.rs
```diff
@@ -222,28 +222,39 @@ pub fn set_decoration_mode(
...
-    let Some(output) = tag.output(&state.pinnacle) else {
+    let Some(target_output) = tag.output(&state.pinnacle) else {
         return;
     };
 
-    state.pinnacle.begin_layout_transaction(&output);
-    state.pinnacle.request_layout(&output);
+    if let Some(source_output) = source_output {
+        if source_output != target_output {
+            state.pinnacle.begin_layout_transaction(&target_output);
```
Ottatop: We should also take snapshots for the target output here.

Ottatop (changes requested):

src/api/window.rs
anriha: I would guess that `is_window_mapped` would update after the layout transaction complete. So I am thinking that maybe this logic should be different. `is_window_mapped` would be used to check if we should update the `source_output`. And for `target_output` we should check that the tag is active?

src/api/window.rs
anriha: Looking at it, wouldn't it make more sense here to use tag output instead of window output? The window could be on multiple outputs at the same time, and we could get the incorrect output. Overall the API for `window.output` is probably incorrect in this way. There is no guarantee that window is only on one output. 

But to the original point, the issue is that the borrow checker won't allow me to do `Some(target_output)` so I am not sure if this is not easier.

src/api/window.rs
Ottatop: `is_window_mapped` checks if a window's buffer is mapped (or if `set_mapped(true)` has been called on an `X11Surface`). It should be true shortly after window rules are applied. The check is used to prevent the layout from triggering when window rules are still executing, because if a layout finishes before window rules do then it will trigger these asserts: https://github.com/pinnacle-comp/pinnacle/blob/451bf36cf2a557dc823128983c01a6abbbaa89f4/src/api/window/v1.rs#L477-L489 So apart from triggering possibly useless layouts, it may also crash the compositor, so it needs to be moved after the check.

> And for `target_output` we should check that the tag is active?

Yea, we could do that.

src/api/window.rs
Ottatop: The plan is to disallow windows from having tags from multiple outputs. I'm currently writing the wiki for the project and I've defined setting multiple tags from different outputs on a window as undefined behavior. I just haven't gotten around to enforcing that in the code.

> But to the original point, the issue is that the borrow checker won't allow me to do `Some(target_output)` so I am not sure if this is not easier.

Ah, didn't look at the exact types. You can do `if source_output.as_ref() != Some(&target_output)`.

src/api/window.rs
anriha: Thanks for the explanation. I changed it and added check for active tag.

src/api/window.rs
anriha: Makes sense, especially if it is also part of the API. Thanks for the pointer, I changed it.

src/api/window.rs
anriha: Done

Ottatop (approved):

Ottatop:
LGTM, thanks!

PR #302, opened 2025-03-29
release: Prepare for 0.1.0-alpha.1

Ottatop:
Well if I don't put a number to the name then we're never gonna get some sort of version, so here we are! I'm pretty comfortable with where we are right now regarding bugs n stuff, big thanks to @ildar for reporting a bunch of issues

Still need to figure out the CI workflow for this and what stuff needs to go into the release. Wiki should be versioned now, but docs still need to be generated by a release workflow.


Ottatop:
And now I have the pleasure of debugging the workflow when it inevitably fails ☺️

Issue #314, opened 2025-05-21
Make snowcap::wgpu use WGPU environment variable configurations.


sapient-cogbag:
Currently, `pinnacle` (or more specifically, `snowcap`) ignores the [WGPU environment variables](https://crates.io/crates/wgpu#environment-variables) specifically for selecting an adapter (by setting [instance flags](https://docs.rs/wgpu/25.0.0/wgpu/struct.InstanceFlags.html#method.with_env)), including specifically the `WGPU_ALLOW_UNDERLYING_NONCOMPLIANT_ADAPTER` variable (which I need to use because I'm on very old hardware which does not have a vulkan implementation marked as complete, and hence I want to do this to see how well it works in practice to switch from Awesome to Pinnacle at some point ^.^).

This issue is essentially requesting that the functionality for using the relevant envvars be enabled - it is a one line change and one that I will do myself so it can be merged if desired or not :p

Ottatop:
Yea that sounds like a good idea, PR welcomed.

sapient-cogbag:
Implemented by #315, closing ^.^

PR #315, opened 2025-05-21
feature: enable use of wgpu instance flag environment variables

sapient-cogbag:
Makes `pinnacle` (and specifically `snowcap`) use the WGPU environment variables for modifying [`InstanceFlags`](https://docs.rs/wgpu/25.0.0/wgpu/struct.InstanceFlags.html#method.with_env).

Should close issue #314 :3


Ottatop (approved):

Ottatop:
lgtm, thanks!

PR #322, opened 2025-06-28
justfile: Fix various error when building from source

Ph4ntomas:
This PR fixes various issue I encountered while building from main.
- `just` invocation failed if started without `LUA_CPATH` in the environment
- `just install` failed due some dependencies failing to build
- `just clean` complains about trying to remove `lua-grpc-client` even though the rock no longer exists

> `just install` failed due some dependencies failing to build

This is outside of pinnacle control. `luaossl`, a dependency of `http`, fails to build on recent compilers due to some breaking changes in recent C syntax.
However, the way luarocks is invoked did not allow to search for dependencies in the system wide tree, so I changed the behavior.

The last commit of the series could be removed. I uses luarocks to retrieve the paths when available, instead of depending on a hardcoded string + LUA_{C,}PATH, which already does the correct thing

### DIFF
--- a/justfile
+++ b/justfile
@@ -34,7 +34,6 @@ install-lua-lib: gen-lua-pb-defs
 clean: clean-snowcap
     rm -rf "{{xdg_data_dir}}"
     -luarocks remove --local pinnacle-api
-    -luarocks remove --local lua-grpc-client
 
 # Run `cargo build`
 build *args: gen-lua-pb-defs
--- a/justfile
+++ b/justfile
@@ -28,7 +28,7 @@ install-protos:
 install-lua-lib: gen-lua-pb-defs
     #!/usr/bin/env bash
     cd "{{rootdir}}/api/lua"
-    luarocks make --local --lua-version "{{lua_version}}" pinnacle-api-dev-1.rockspec
+    luarocks make --local --deps-mode "order" --lua-version "{{lua_version}}" pinnacle-api-dev-1.rockspec
 
 # Remove installed configs and the Lua API (requires Luarocks)
 clean: clean-snowcap
--- a/justfile
+++ b/justfile
@@ -7,7 +7,7 @@ lua_version := "5.4"
 
 local_lua_path := x"$HOME/.luarocks/share/lua/" + lua_version + x"/?.lua;$HOME/.luarocks/share/lua/" + lua_version + "/?.init.lua;"
 export LUA_PATH := local_lua_path + env("LUA_PATH", "")
-export LUA_CPATH := x"$HOME/.luarocks/lib/lua/" + lua_version + x"/?.so;$LUA_CPATH"
+export LUA_CPATH := x"$HOME/.luarocks/lib/lua/" + lua_version + x"/?.so;" + env("LUA_CPATH", "")
 
 list:
     @just --list --unsorted
--- a/justfile
+++ b/justfile
@@ -5,9 +5,24 @@ xdg_data_dir := `echo "${XDG_DATA_HOME:-$HOME/.local/share}/pinnacle"`
 
 lua_version := "5.4"
 
+# dirty trick until just's which() is stabilized.
+luarocks := `which luarocks 2>/dev/null || true`
+luarocks_full := if luarocks != "" { `luarocks path --help | grep -- "--full " || true`  } else { "" }
+
 local_lua_path := x"$HOME/.luarocks/share/lua/" + lua_version + x"/?.lua;$HOME/.luarocks/share/lua/" + lua_version + "/?.init.lua;"
-export LUA_PATH := local_lua_path + env("LUA_PATH", "")
-export LUA_CPATH := x"$HOME/.luarocks/lib/lua/" + lua_version + x"/?.so;" + env("LUA_CPATH", "")
+local_lua_cpath := x"$HOME/.luarocks/lib/lua/" + lua_version + x"/?.so;"
+
+export LUA_PATH := if luarocks_full != "" {
+    shell(luarocks + ' "$@"', 'path', '--full', '--lr-path', '--lua-version', lua_version)
+} else if luarocks != "" {
+    shell(luarocks + ' "$@"', 'path', '--lr-path', '--lua-version', lua_version) + ";" + env("LUA_PATH", "")
+} else { local_lua_path + env("LUA_PATH", "") }
+
+export LUA_CPATH := if luarocks_full != "" {
+    shell(luarocks + ' "$@"', 'path', '--full', '--lr-cpath', '--lua-version', lua_version)
+} else if luarocks != "" {
+    shell(luarocks + ' "$@"', 'path', '--lr-cpath', '--lua-version', lua_version) + ";" + env("LUA_CPATH", "")
+} else { local_lua_cpath + env("LUA_CPATH", "") }
 
 list:
     @just --list --unsorted
--- a/snowcap/justfile
+++ b/snowcap/justfile
@@ -21,7 +21,6 @@ install-protos:
 clean:
     rm -rf "{{xdg_data_dir}}"
     -luarocks remove --local snowcap-api
-    -luarocks remove --local lua-grpc-client
 
 # Generate the protobuf definitions Lua file
 gen-lua-pb-defs:


Ph4ntomas:
Huh, I think I forgot to export luarocks path. I'll see if I can do it from the just file tho

Ph4ntomas:
Huh,  `luarocks path --full` was added in `luarocks` 3.10, the github action pulls 3.8.0 :(
Let me fix this

Ottatop:
If you rebase I bumped the luarocks action up to 5 for luarocks 3.11, was in the middle of reviewing

Ph4ntomas:
Oh thanks ! then I'll just rebase :D

Ph4ntomas:
I've been thinking about the CI issue. While the CI should pass with the rebase, it also put a hard dependency on luarocks 3.10, which I'm not super fan of, since lots of distros still ship 3.8.0.

I'm going to change the logic so luarocks is used only if the LUA_ environment var are empty. That way there is a fallback if the luarock version is too low

Ph4ntomas:
> I'm going to change the logic so luarocks is used only if the LUA_ environment var are empty. That way there is a fallback if the luarock version is too low

In addition to that, I've added a more descriptive error if LUA_* are not set and luarocks doesn't support the `--full` option.

justfile
```diff
@@ -5,9 +5,26 @@ xdg_data_dir := `echo "${XDG_DATA_HOME:-$HOME/.local/share}/pinnacle"`
...
+lr_cpath := if luarocks_full != "" { shell(luarocks + ' "$@"', 'path', '--full', '--lr-cpath', '--lua-version', lua_version) } else { "" }
+
+env_lua_path := env("LUA_PATH", "")
+env_lua_cpath := env("LUA_CPATH", "")
+
+export LUA_PATH := if env_lua_path != "" {
+    x"$HOME/.luarocks/share/lua/" + lua_version + x"/?.lua;$HOME/.luarocks/share/lua/" + lua_version + "/?.init.lua;" + env_lua_path
+} else if lr_path != "" {
+    lr_path
+} else { error("LUA_PATH not defined, and luarocks is either missing or < 3.10.0") }
```
Ottatop: The `error` prevents users who only want to use the Rust API from building, I'd try to downgrade it to a print or something and default to `env_lua_path`. Plus I don't think we need to error in the case where luarocks isn't installed because it's required for local installs of the Lua API.

justfile
```diff
@@ -5,9 +5,26 @@ xdg_data_dir := `echo "${XDG_DATA_HOME:-$HOME/.local/share}/pinnacle"`
...
+lr_path := if luarocks_full != "" { shell(luarocks + ' "$@"', 'path', '--full', '--lr-path', '--lua-version', lua_version) } else { "" }
+lr_cpath := if luarocks_full != "" { shell(luarocks + ' "$@"', 'path', '--full', '--lr-cpath', '--lua-version', lua_version) } else { "" }
+
+env_lua_path := env("LUA_PATH", "")
+env_lua_cpath := env("LUA_CPATH", "")
+
+export LUA_PATH := if env_lua_path != "" {
+    x"$HOME/.luarocks/share/lua/" + lua_version + x"/?.lua;$HOME/.luarocks/share/lua/" + lua_version + "/?.init.lua;" + env_lua_path
+} else if lr_path != "" {
+    lr_path
```
Ottatop: From cursory investigation, I believe `luarocks path --full --lr-path` will append the contents of `LUA_PATH` to the end, so I believe we can combine the cases where `LUA_PATH` exists and `lr_path` exists to just return `lr_path`. I suppose in the case that luarocks is too old we could just use the non-full lr-path and append `env_lua_path` manually?

justfile
```diff
@@ -5,9 +5,26 @@ xdg_data_dir := `echo "${XDG_DATA_HOME:-$HOME/.local/share}/pinnacle"`
...
+} else { error("LUA_PATH not defined, and luarocks is either missing or < 3.10.0") }
+
+export LUA_CPATH := if env_lua_cpath != "" {
+    x"$HOME/.luarocks/lib/lua/" + lua_version + x"/?.so;" + env_lua_cpath
+} else if lr_cpath != "" {
+    lr_cpath
+} else { error("LUA_CPATH not defined, and luarocks is either missing or < 3.10.0") }
```
Ottatop: Similar stuff to the above here

justfile
Ph4ntomas: Ha ! So I managed to reintroduce the bug that started this PR ...

justfile
Ph4ntomas: > From cursory investigation, I believe luarocks path --full --lr-path will append the contents of LUA_PATH to the end, so I believe we can combine the cases where LUA_PATH exists and lr_path exists to just return lr_path. 

Technically, with `path --full --lr-path` option, luarocks appends the content of lua's `package.path` variable, which is either LUA_PATH or a hardcoded default system path.

>  I suppose in the case that luarocks is too old we could just use the non-full lr-path and append env_lua_path manually?

Agreed. The alternative to correctly simulate `--full` would be to either parse `luarocks path` output to retrieve the  string, or use lua<version> to print the package.  
Both option are a bit too convoluted for my taste, especially when `eval $(luarocks path)` can be run by the user ahead of time.

From the other conversation:
> I'd try to downgrade it to a print or something and default to env_lua_path

There is no warning facility in `just` toplevel. The closest thing we can do when the environment is empty, is to retrieve it like this:
```
env("LUA_PATH", `echo "Warning: LUA_PATH is not set" >&2 && echo ""`)
```

This is a huge hack so I did not include it, but let me know if you feel we should warn the user when we fallback on `luarocks path --lr-path` while the env is unset

Ph4ntomas:
I'm not sure what's your policy on resolving conversation, so I left them open, but I've addressed them in the latest push

justfile
Ottatop: > There is no warning facility in `just` toplevel.

Ah, unlucky. Yea that echo is pretty hacky, I think it's fine to not print any warnings.

Ottatop:
Definitely liking this a lot more than the ad-hoc stuff I was doing before, thank you!

> I'm not sure what's your policy on resolving conversation

Don't really have any policy, I guess resolve them if you feel you've addressed everything

justfile
```diff
@@ -5,9 +5,24 @@ xdg_data_dir := `echo "${XDG_DATA_HOME:-$HOME/.local/share}/pinnacle"`
 
 lua_version := "5.4"
 
+# dirty trick until just's which() is stabilized.
+luarocks := `which luarocks 2>/dev/null || true`
+luarocks_full := if luarocks != "" { `luarocks path --help | grep -- "--full " || true`  } else { "" }
```
ildar, line 10: `if luarocks != ""` always true

justfile
```diff
@@ -28,13 +43,12 @@ install-protos:
 install-lua-lib: gen-lua-pb-defs
     #!/usr/bin/env bash
     cd "{{rootdir}}/api/lua"
-    luarocks make --local --lua-version "{{lua_version}}" pinnacle-api-dev-1.rockspec
+    luarocks make --local --deps-mode "order" --lua-version "{{lua_version}}" pinnacle-api-dev-1.rockspec
```
ildar, line 46: if the `luarocks` variable is defined in the line 9 then it could (should?) be used here

justfile
Ph4ntomas: No, but I can see why it's confusing.
- `luarocks :=`: This define a `just` variable, that contains the (standard) output of what's on the right hand side of `:=`
-  `which luarocks 2>/dev/null || true`: This execute `which luarocks` silencing any errors. `|| true` is there because `just` stops if a shell command fail. This will either output luarock path, or nothing.
- `if luarocks != ""`: If `luarocks` is installed, the variable contains it's path. If `luarocks` is not installed, the variable is empty, so the condition is false.

justfile
Ph4ntomas: Yeah, it could. I don't think that would have any impact, except on the off chance someone override the PATH variable via a global export/unexport (but then you'd have issues with every other rules)

justfile
ildar: you're right here in case luarocks is `/.../bin/luarocks`. But if you try to use it like:
```
just --set luarocks /usr/bin/luarocks-5.4 install
```
then it's totally wrong.

ildar:
and yes, actually having `PREFIX` or the `DEST_DIR` would be most desirable.

justfile
Ph4ntomas: > then it's totally wrong.

How so ? If you explicitly sets the variable, the condition is true, but that doesn't make it "always true".
Plus the goal of the PR was to fix the default behavior, so this is kinda out of scope.

Did you meant to answer the conversation about using {{luarocks}} in the install rule ? (It would make more sens in that context)

In any case if you think there are still improvements to be made, IMO you should open an issue or a new PR, since this one has been closed a week ago. While I still receive mails for it, it's not as visible as it could be.

ildar: > this one has been closed a week ago

True. I am late with this PR, I understand it clearly. I wrote the above just For X's Information in case `justfile` will be reworked again.

Issue #237, opened 2024-05-10
Add floating layout


Ottatop:
The idea is simple:

1. Add a "pseudo-floating" state to windows to denote windows that are supposed to be tiled but due to a lower-than-requested amount of geometries from a layout generator are floating
2. Add a layout generator that does literally nothing
3. ???
4. Profit

ildar:
workaround, just FTR:
```
    Window.add_window_rule(function(window)
        window:set_floating(true)
        end)
```

Ottatop:
Implemented in #326

PR #326, opened 2025-07-05
Add floating layout

Ph4ntomas:
I took a jab at #237.

The code needs cleanup, and there are some tests failing, but the feature itself is mostly working.

From #237:
>   1.  Add a "pseudo-floating" state to windows to denote windows that are supposed to be tiled but due to a lower-than-requested amount of geometries from a layout generator are floating

I called this state `Spilled` because I found the image funny, but I'm open to suggestions

>  2.  Add a layout generator that does literally nothing
Not in this PR at the moment

>  3.  ???

The way it's currently implemented is as follow:
When we compute the layout, we count Tiled and Spilled window as the amount of nodes requested.
When we apply the layout to window, we collect windows which did not fit the tree.
We apply the 'Spilled' set to them. This allow other part of the compositor to apply the same 'rule' as for floating window.

If however all window fit (tiled + spilled), the spilled window gets set back to the 'Tiled' layout mode. This is done to re-tiled then when switching layouts.


src/layout.rs
```diff
@@ -164,6 +176,12 @@ impl Pinnacle {
 
         for (win, geo, is_tiled) in wins_and_geos.iter() {
             if *is_tiled {
+                // Just remove the spilled here if the window has been tiled.
+                // Since this will not change the layout, it shouldn't matter that we don't use a
+                // transacion.
+                //
+                // TODO: confirm with Check with ottatop
```
Ph4ntomas: @Ottatop what are you thought on this matter ?

src/layout.rs
```diff
@@ -123,16 +123,28 @@ impl Pinnacle {
...
             warn!("Cannot update_windows_from_tree without output geo");
-            return;
+            return Vec::new();
         };
 
         let non_exclusive_geo = layer_map_for_output(output).non_exclusive_zone();
 
+        // TODO: I'm not a huge fan of doing it that way, but until
+        // State.update_window_layout_mode_and_layout gets refactored, it's best to defer setting
+        // the layout since we also need to compute the window geometry. This is a POC and I'm lazy
```
Ph4ntomas: This feels like a hack, and it very much is, but the only way to call the function to properly handling switching to a pseudo floating mode without issue was to return the vector of spilled windows to the calling scope. 

That being said, I think we could get away with just setting the Spilled layout mode here. Since it acts like a floating behavior, we don't need re-compute the layout anyway, so we might not need the full transaction handling logic from that scope.

If that's true, maybe we can just extract the logic to compute the new geometry form State.update_window_layout_mode_and_layout.

Any thought on this ?

src/grab/resize_grab.rs
```diff
@@ -401,7 +401,9 @@ impl State {
             };
 
             if window.with_state(|state| {
-                !state.layout_mode.is_floating() && !state.layout_mode.is_tiled()
+                !state.layout_mode.is_floating()
+                    && !state.layout_mode.is_tiled()
+                    && !state.layout_mode.is_spilled()
```
Ottatop: Probably should just make this `state.layout_mode.is_maximized() || state.layout_mode.is_fullscreen()`

src/grab/resize_grab.rs
```diff
@@ -451,9 +453,11 @@ impl State {
             return;
         };
 
-        if window
-            .with_state(|state| !state.layout_mode.is_floating() && !state.layout_mode.is_tiled())
-        {
+        if window.with_state(|state| {
+            !state.layout_mode.is_floating()
+                && !state.layout_mode.is_tiled()
+                && !state.layout_mode.is_spilled()
```
Ottatop: Same here

src/layout.rs
Ottatop: Should be fine, they'll get picked up by the transaction for tiled windows below

src/layout.rs
Ph4ntomas: > so we might not need the full transaction handling logic from that scope.

This might not hold for window that were previously floated, since pinnacle reuse the previous value.

On the other hand, the function we're in already create a transaction since window may shift and be resized. I'll try to compute new layouts from the function current next

Ottatop: Yes, transactions are needed to move and resize the windows to their possibly-different stored geometries.

I think it's a good idea to extract the floating update logic and use it on the spilled windows. This way we can piggyback their updates onto the tiled transaction below and only have to deal with 1 transaction vs a bunch of small ones per spilled window.

src/grab/resize_grab.rs
Ph4ntomas: At some point, I was considering adding a "is_resizeable()", which would act as a pseudo layout in this context, while being clearer than a bunch of boolean. 

I kept that discussion for the PR tho.

I can go with the simplified boolean logic or move that a function. Which option do you prefer ?

src/grab/resize_grab.rs
Ph4ntomas: Agreed, I'm closing this one to keep the discussion above

src/grab/resize_grab.rs
Ottatop: If it's only going to be used for just these two boolean exprs then I think the simple boolean logic is better, plus we'd still probably have to check individual conditions if we use `is_resizable` in other contexts.

Ph4ntomas:
> Ok, initial draft looks pretty good 👍

Cool, thanks ! :D

> A few things outside of the review comments:
> 
>     1. Currently, dragging a spilled window onto another output doesn't move it to that output. We should use the logic for [tiled move grabs across outputs](https://github.com/pinnacle-comp/pinnacle/blob/4d470117df4a905cc76620fda9a06c242ce34227/src/grab/move_grab.rs#L71-L82) for spilled windows to update their tags and layout the new output.
> 
>     2. I believe `WindowHandle::set_geometry` will be able to move spilled windows onto different outputs without changing its tags. We could include spilled windows in the floating window tag update logic [here](https://github.com/pinnacle-comp/pinnacle/blob/4d470117df4a905cc76620fda9a06c242ce34227/src/window.rs#L415-L443), but that clashes directly with the point above, not sure how to handle that at the moment.

Yeah IMO the 'floating' behavior should only be on the current tag/output pair. I'll check what I can do wrt 1 first.

Tbh I'm still running pinnacle through winit since I want to keep access to the files I'm editing even if it crashed, so I haven't touched the multi head behavior just yet. I'll address the other stuff, then come back to that point.
 
>     3. You'll also need to add `WindowHandle:is_spilled` to the Lua API.

Will do !

>     4. While I _do_ like using "spilled" as the term here, it's not that self-explanatory for people using the API. Unfortunately I haven't come up with anything better.

When I started, it was briefly called `AutoFloating` but I found that even more confusing since that's what I would expect for e.g. popup window. Personally I feel like `Spilled`, even if not necessarily obvious is easy to remember, so updating the wiki/API documentation might be enough if we keep that name.

Since this is a rather small detail, I'll address the other points first though. If anything better comes up, I can always change this before this gets merged.

src/layout.rs
Ph4ntomas: Done with [a96d222](https://github.com/pinnacle-comp/pinnacle/pull/326/commits/a96d2223f220ae66c3eb5f428b1663f901ef9075)

src/grab/resize_grab.rs
Ph4ntomas: Done with [5abb634](https://github.com/pinnacle-comp/pinnacle/pull/326/commits/5abb634dc4cdac3fd26d481dcf5571284185e4d6)

Ph4ntomas:
@Ottatop I think I've addressed everything.

[e1c2f0c](https://github.com/pinnacle-comp/pinnacle/pull/326/commits/e1c2f0c8a0f4d620ff4a58dd57ca6b8e92da122b) has two changes:
- one related to moving a tiled|spilled window to another monitor, so the behavior is well defined (address the quirks I've mentioned previously.)
- one related to swapping a tiled window with a spilled one

> Check what can be done to swap a spilled window with a tiled one ?

I briefly explored this by applying the same swapping behavior in use in tiled window movement, but since the swap occur on hover, you could swap with a random while dragging, which is less than ideal.
This can be mitigated by limiting the region being considered to a fraction of the window you're hovering (e.g. the tiltle bar, or the bottom n percent/pixels of the target window, but it still means you can swap involuntarily, make the lower window spilled, and mess with your whole layout. In my opinion, it's better not to implement this, or at least not with a "swap on hover" behavior.

An alternative behavior would be to check for a window when releasing the grab, and swap only then, but I don't want to invest more time in this behavior if you don't want it altogether. This would come with the same caveat that you'd want a smaller region as a detection zone to prevent randomly swapping when you're done moving your spilled window.
 
Regardless, it's already possible to perform the swap by starting with a tiled window, and this behavior is more consistent.

src/window.rs
```diff
@@ -440,6 +446,73 @@ impl Pinnacle {
             }
         }
     }
+
+    pub fn compute_window_geometry(
+        &self,
+        window: &WindowElement,
+        output: &Output,
+        output_geo: &Rectangle<i32, Logical>,
+    ) -> Option<Rectangle<i32, Logical>> {
```
Ottatop: Instead of passing in both the output and its geometry, you can either:
1. Pass in just the output and derive the geometries from it alone, or
2. Pass in the output geometry along with the non-exclusive geometry instead of the output.

I'm indifferent to which option you choose. If you go with 2, change the `&Rectangle<..>` to just `Rectangle<..>`, rectangles are `Copy`.

src/layout.rs
```diff
@@ -120,11 +120,18 @@ impl Pinnacle {
...
-            .cloned();
+            .filter(|win| {
+                win.with_state(|state| {
+                    state.layout_mode.is_tiled() || state.layout_mode.is_spilled()
+                })
+            })
+            .cloned()
+            .partition::<Vec<_>, _>(|win| win.with_state(|state| state.layout_mode.is_tiled()));
+
+        let tiled_windows = tmp_tiled_windows.into_iter().chain(tmp_spilled_windows);
```
Ottatop: The problem with decoupling geometry assignment from window order is that it can cause later layouts to visually reorder windows. Coupling this with the change that causes tiled windows moved to other outputs to be briefly spilled makes the issue more pronounced.

For example, if I have two outputs, one with an empty layout generator and one with master stack, I can drag a spilled window onto the master stack output and it will tile at the end (easier to reproduce if there's a layer shell bar that prevents the window from swapping). Then, I can open another window and that previous window will re-tile at its actual window order, causing it to jump somewhere unexpected.

Honestly, I think I'm fine with keeping the previous behavior where you could "kick out" one of the tiled windows at the end. The idea of spilled windows only has two use cases imo:
1. A floating layout, where this isn't an issue, and
2. To let windows still show up for people making their own layout generators with incorrect numbers of nodes.

I think deliberately supporting something like a layout generator with a capped amount of windows is out of scope.

src/layout.rs
```diff
@@ -230,6 +251,12 @@ impl Pinnacle {
         let win2_index = self.windows.iter().position(|win| win == win2);
 
         if let (Some(first), Some(second)) = (win1_index, win2_index) {
+            let win1_mode = win1.with_state(|s| s.layout_mode);
+            let win2_mode = win2.with_state(|s| s.layout_mode);
+
+            win1.with_state_mut(|s| s.layout_mode = win2_mode);
+            win2.with_state_mut(|s| s.layout_mode = win1_mode);
```
Ottatop: Is this needed to let something work?

src/grab/move_grab.rs
```diff
@@ -66,8 +66,24 @@ impl PointerGrab<State> for MoveSurfaceGrab {
...
         match layout_mode {
             LayoutModeKind::Tiled => {
+                // We've just switched ouput, assume spill. If the window fit, the layout
+                // request will fix it. If not, then the behavior will be sane.
+                // It also have the added benefit appending to the target tree, instead of
+                // keeping the current window order (and thus being somewhat random)
+                if win_output != other_output {
+                    self.window
+                        .with_state_mut(|s| s.layout_mode.set_spilled(true));
+                }
```
Ottatop: I'm not a fan of initially spilling moved tiled windows, see my comment in `src/layout.rs` for why

src/grab/move_grab.rs
```diff
@@ -66,8 +66,24 @@ impl PointerGrab<State> for MoveSurfaceGrab {
             .as_ref()
             .and_then(|op| op.upgrade());
 
+        let win_output = self.window.output(&state.pinnacle).map(|o| o.name());
+        let other_output = output_under_pointer.as_ref().map(|o| o.name());
```
Ottatop: `Output`s are ref-counted and `PartialEq`, no need to map to their name to compare

Ottatop (commented):
> I briefly explored this by applying the same swapping behavior in use in tiled window movement, but since the swap occur on hover, you could swap with a random window while dragging, which is less than ideal.
This can be mitigated by limiting the region being considered to a fraction of the window you're hovering (e.g. the title bar, or the bottom n percent/pixels), but it still means you can swap involuntarily, make the lower window spilled, and mess with your whole layout. In my opinion, it's better not to implement this, or at least not with a "swap on hover" behavior.

I don't think we should be allowing swaps between spilled and tiled windows. Actually, looking at the swap logic, it's only preventing swaps with floating windows when it should probably prevent swaps with any window that isn't tiled.

api/lua/pinnacle/window.lua
```diff
@@ -651,6 +651,16 @@ function WindowHandle:tiled()
     return response and response.layout_mode == layout_mode_def.LAYOUT_MODE_TILED or false
 end
 
+---Gets whether this window was spilled by the layout.
```
Ottatop:
```suggestion
---Gets whether this window is spilled from the layout.
---
---A window is spilled whenever the current layout generator does not contain enough nodes to tile it.
---It acts like a floating window, but will tile once a layout generator returns a layout tree with enough nodes.
```

api/rust/src/window.rs
```diff
@@ -677,6 +680,26 @@ impl WindowHandle {
...
+    pub fn tiled(&self) -> bool {
+        self.tiled_async().block_on_tokio()
+    }
+
+    /// Async impl for [`Self::tiled`].
+    pub async fn tiled_async(&self) -> bool {
+        self.layout_mode_async().await == LayoutMode::Tiled
+    }
+
+    /// Gets whether or not this window was spilled by the layout.
```
```suggestion
    /// Gets whether or not this window is spilled from the layout.
    ///
    /// See [`LayoutMode::Spilled`] for more information.
```

api/rust/src/window.rs
```diff
@@ -180,6 +180,8 @@ pub enum LayoutMode {
     Fullscreen,
     /// The window is maximized.
     Maximized,
+    /// The window was spilled from the layout
```
```suggestion
    /// The window is spilled from the layout.
    /// 
    /// A window is spilled whenever the current layout generator does not contain enough nodes to tile it.
    /// It acts like a floating window, but will tile once a layout generator returns a layout tree with enough nodes.
```

api/rust/src/window.rs
```diff
@@ -677,6 +680,26 @@ impl WindowHandle {
         self.layout_mode_async().await == LayoutMode::Floating
     }
 
+    /// Gets whether or not this window is tiled
```
```suggestion
    /// Gets whether or not this window is tiled.
```

Ottatop (changes requested):
> Personally I feel like `Spilled`, even if not necessarily obvious is easy to remember, so updating the wiki/API documentation might be enough if we keep that name.

Yea I think this is the best course of action, I can't think of anything more self-explanatory.

src/layout.rs
Ph4ntomas: Yeah, but that wont be necessary anymore. It's there to allow swapping a tiled window with a spilled one, which will not be supported explicitely anymore

src/layout.rs
Ph4ntomas: I thought about the re ordering issue, then didn't add it in because I did not notice the random reordering.

I think I mostly targetted a specific tile when moving from a screen to the other which fixes the behavior by doing an explicit swap, then I forgot about it, good thing you noticed it then.

> I think deliberately supporting something like a layout generator with a capped amount of windows is out of scope.

Well that remove the very issue I was trying fix. And it simplify a lot of things (no need to spill the window before moving, no need to change the swap)

src/grab/move_grab.rs
Ph4ntomas: Will be removed, since it was only necessary for a sane behavior when going from a tilled layout to one with a capped number of node

src/grab/move_grab.rs
Ph4ntomas: Woops, I added these for logging purposed reused them for the comparison ...

Obviously this'll go

src/layout.rs
Ph4ntomas: For the record:

> Actually, looking at the swap logic, it's only preventing swaps with floating windows when it should probably prevent swaps with any window that isn't tiled.

I purposefully allowed swapping a tiled window and a spilled on, and this was to make the swap works.

Since explicitely supporting capped layout will not be kept, this is no longer needed, so I'll additionally prevent swapping a tiled window with a spilled one.

src/window.rs
Ph4ntomas: I'm going with two then.

The reason I was passing the output & its geometry was make this function infallible. In my opinion, the function should only care about window geometry, and `None` is a valid value (as in, the window doesn't have its own geometry, so it takes the 'parent' geometry).

Going with 1, this would change to None pulling double duty of both meaning `the output is un-mapped` and the previously mentioned valid value.

api/rust/src/window.rs
Ph4ntomas: Done + squashed in next push.

api/rust/src/window.rs
Ph4ntomas: Done + squash in next push

api/rust/src/window.rs
Ph4ntomas: Done + squashed in next push

src/layout.rs
Ph4ntomas: > I think deliberately supporting something like a layout generator with a capped amount of windows is out of scope.

For the record, this *could* be done by reordering window such that:
- the 'sort' is stable
- window that are just moved in come last
- spilled window are always last

However I've removed that for the time being.

Ph4ntomas:
I've applied the requested changes. 

I'll rebase the PR to squash the `-rework` items, but I want to confirm the correctness of the behavior/implementation first.

Ph4ntomas:
@Ottatop let me know where to document the `Spilled` layout.

I'd guess the wiki would be a good place, but I don't see anything related to window LayoutMode.
Maybe I should just expand the api in-code documentation

For the lua generator, even though the floating layout does not *need* to be wrapped in a function, I did it so the API is consistent with the other builting generators

api/lua/pinnacle/layout.lua
```diff
@@ -644,6 +644,26 @@ function builtin.fair(options)
     }
 end
 
+---A layout generator that leaves windows in a pseudo floating state.
+---The window are left in the spilled state. They act as if in a floating
+---state, except they become tiled again if the layout changes to one
+---providing enough nodes.
```
Ottatop: I don't think we need to elaborate on the behavior of spilled windows in the floating generator docs. I'd change this to something like
```
---A layout generator that floats windows.
---
---This works by simply returning an empty layout tree.
---Note: the windows are not truly floating, see `WindowHandle::is_spilled` for details.
```

api/lua/pinnacle/layout.lua
```diff
@@ -644,6 +644,26 @@ function builtin.fair(options)
...
+---@class pinnacle.layout.builtin.Floating : pinnacle.layout.LayoutGenerator
+
+---Creates a layout generator that leaves window in a pseudo floating state.
```
Ottatop:
```suggestion
---Creates a layout generator that floats windows.
```

api/rust/src/layout/generators.rs
```diff
@@ -531,6 +531,22 @@ impl LayoutGenerator for Fair {
     }
 }
 
+/// A [`LayoutGenerator`] that leaves windows in a pseudo floating state.
+///
+/// The window are left in the [`Spilled`] state. They act as if in a floating
+/// state, except they become tiled again if the layout changes to one
+/// providing enough nodes.
+///
+/// [`Spilled`]: crate::window::LayoutMode::Spilled
```
Ottatop: See comment on Lua docs

Ottatop (commented):
> let me know where to document the `Spilled` layout.

Layout modes are documented [here](https://github.com/pinnacle-comp/pinnacle/blob/main/wiki/usage/window-management.md) (admittedly not a very self-evident page name).

> Maybe I should just expand the api in-code documentation

Probably wouldn't hurt to do that as well, or at least something short if you're adding it to the wiki.

> For the lua generator, even though the floating layout does not need to be wrapped in a function, I did it so the API is consistent with the other builting generators

This is good, makes it more discoverable

api/lua/pinnacle/layout.lua
Ph4ntomas: > see `WindowHandle::is_spilled` for details.

I've added some details regarding the behavior on the `WindowHandle::spilled` function

api/lua/pinnacle/window.lua
```diff
@@ -651,6 +651,21 @@ function WindowHandle:tiled()
     return response and response.layout_mode == layout_mode_def.LAYOUT_MODE_TILED or false
 end
 
+---Gets whether this window is spilled from the layout.
+---
+---A Window is spilled when the current layout doesn't contains enough nodes
+---and the compositor cannot assign a geometry to it. In that state, the window
+---behave as a floating window except that it gets tiled again if the number of
```
Ottatop:
```suggestion
---behaves as a floating window except that it gets tiled again if the number of
```

api/lua/pinnacle/window.lua
```diff
@@ -651,6 +651,21 @@ function WindowHandle:tiled()
     return response and response.layout_mode == layout_mode_def.LAYOUT_MODE_TILED or false
 end
 
+---Gets whether this window is spilled from the layout.
+---
+---A Window is spilled when the current layout doesn't contains enough nodes
```
Ottatop:
```suggestion
---A window is spilled when the current layout doesn't contains enough nodes
```

api/rust/src/window.rs
```diff
@@ -677,6 +680,30 @@ impl WindowHandle {
...
+    }
+
+    /// Async impl for [`Self::tiled`].
+    pub async fn tiled_async(&self) -> bool {
+        self.layout_mode_async().await == LayoutMode::Tiled
+    }
+
+    /// Gets whether or not this window is spilled from the layout.
+    ///
+    /// A Window is spilled when the current layout doesn't contains enough nodes and the
```
```suggestion
    /// A window is spilled when the current layout doesn't contains enough nodes and the
```

api/rust/src/window.rs
```diff
@@ -677,6 +680,30 @@ impl WindowHandle {
...
+    /// compositor cannot assign a geometry to it. In that state, the window behave as a floating
```
```suggestion
    /// compositor cannot assign a geometry to it. In that state, the window behaves as a floating
```

wiki/usage/window-management.md
```diff
@@ -31,3 +31,10 @@ Fullscreen windows completely fill the screen. They will also render above all l
 except for those on the overlay layer.
 
 ![Fullscreen layout](/assets/fullscreen.png)
+
+Spilled windows behave like floating windows, but are still tracked by the
+tiled layout. This mode is set by the compositor on window that should be tiled
+but couldn't due to the layout lacking free tiles. Contrary to floating
+windows, spilled ones will become tiled automatically if the layout start
+providing enough empty tiles.
```
Ottatop:
```suggestion
Spilled windows behave like floating windows, but are still tracked by the
tiled layout. This mode is set by the compositor on windows that should be tiled
but aren't due to the layout lacking free tiles. Contrary to floating
windows, spilled ones will become tiled automatically if the layout starts
providing enough empty tiles.
```

Ottatop (changes requested):
Ok this *should* be the final review, everything else lgtm

Issue #330, opened 2025-07-08
Moving fullscreen or maximized window between output doesn't update position


Ph4ntomas:
### Describe the bug

When a floating (or spilled once #326 is merged) window is either fullscreen-ed of maximized, then moved to another output, the window 'floating' position is not updated. This makes the window jumps back to its previous location when the full-screen flag is removed.

For floating window, their tag will be re-updated once they jump back, so that's mostly a weird behavior.

For spilled window, since their tag is currently only updated by mouse movement, they are displayed on one output, but still part of the other, meaning (A is the first monitor, B is where the window was moved while full-screen):
- If a move is initialized on monitor A, the window will be re-attached to that monitor.
- If the active tag on monitor B change to one without the window, it will 'disappear'.
- If monitor B layout changes, the window will jump once more to monitor B.

I would expect a floating, maximized window to become floating again on the monitor it's currently is on, not the one it was before the move.

For spilled window, the same is true, but there should be an additional layout computation, so the window gets properly tiled if it can, or stay spilled but on the current monitor, not the previous one

### Reproduction steps

1. float a window or use the floating layout
2. maximize it
3. move it to another output
4. un-maximize it

### Relevant parts of config

N/A 

### Version info

pinnacle 0.1.0

Build Info:
    Build profile: debug
    rustc version: 1.88.0
    Branch: 237-add-floating-layout
    Commit: 448e70b (wiki: document spilled window)

### Distro

Arch

### Config language

Lua

PR #331, opened 2025-07-09
window: Update position on LayoutMode changes

Ph4ntomas:
When changing LayoutMode, if the window previously had a floating geometry (i.e. it has been floated or spilled before), the window restores it. This lead to some window jumping back to their previous output while being attached to the current output tag.
~~With this fix, we relocate the window on mode change if they do not intersect with current output.~~

~~The intersection check if to restore the previous location for case where the left side of a floating window would be inside another monitor. Without it, you could have 'weird' behavior where toggling fullscreen twice (without moving the window) would move the window.~~

~~This however means that a floating window can still be reattached to its previous output **if** it previously intersected with the one if was fullscreen-ed on.
I find this limitation acceptable, and less confusing than either the previous behavior, or the behavior when we just relocate.~~

Instead of all of the above, the fix just updates the window location when switching output.

Fixes: #330

### DIFF
--- a/src/grab/move_grab.rs
+++ b/src/grab/move_grab.rs
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: GPL-3.0-or-later
 
 use smithay::{
-    desktop::space::SpaceElement,
+    desktop::{layer_map_for_output, space::SpaceElement},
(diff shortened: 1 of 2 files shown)


Ottatop (commented):
Lgtm, thank you!

Ottatop (approved):

PR #335, opened 2025-07-11
Make compositor-initiated window resizing also use single edges

### LINKED ISSUE #162: Make compositor-initiated window resizing also use single edges
Currently resizing uses the two edges closest to the pointer, so you can't resize horizontally without also resizing vertically. The goal is to change this to use one edge if closer to the center of edges and two edges when closer to window corners.

Ph4ntomas:
Previous behavior was checking in which quadrant the pointer was when starting the resize, which
only allowed corner resizing.

~~This patch compute the distance between the cursor location and a fixed set of point.~~

~~The second commit fixes a minor glitch where clicking just outside the main surface would use the shadow as the window location. This had a weird behavior where top and left edges worked as expected, but bottom and right edge had the behavior from the opposing edge.~~

The new behavior is to split the area into a 3x3 grid, and find in which cell the pointer falls into.
The center cell cannot really use that behavior. Instead we use some vector math to determine which the edge to use.

fixes #162

### DIFF
--- a/src/api/window.rs
+++ b/src/api/window.rs
@@ -303,6 +303,10 @@ pub fn resize_grab(state: &mut State, button: u32) {
     let edges = if edges != server::xdg_toplevel::ResizeEdge::None {
         edges
     } else {
+        // Find the closest edge by figuring out which corners the pointer is between.
+        // This works by drawing lines from the window's center to all four corners and the pointer.
+        // Whichever two lines the pointer line is between determines the edge chosen.
+        
         // A bit of an explanation here.
         //
         // The cross product of two vector is `||v1|| * ||v2|| * sin(th)`, with `th` being the
--- a/src/api/window.rs
+++ b/src/api/window.rs
@@ -5,7 +5,7 @@ use smithay::{
     reexports::wayland_protocols::xdg::{
         decoration::zv1::server::zxdg_toplevel_decoration_v1, shell::server,
     },
-    utils::{Point, SERIAL_COUNTER},
+    utils::{Point, SERIAL_COUNTER, Size},
     wayland::seat::WaylandFocus,
 };
 use tracing::warn;
@@ -259,7 +259,7 @@ pub fn resize_grab(state: &mut State, button: u32) {
     else {
         return;
     };
-    let Some((pointer_focus, window_loc)) = state.pinnacle.pointer_contents.focus_under.as_ref()
+    let Some((pointer_focus, _window_loc)) = state.pinnacle.pointer_contents.focus_under.as_ref()
     else {
         return;
     };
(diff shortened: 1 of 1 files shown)


src/api/window.rs
```diff
@@ -270,41 +270,49 @@ pub fn resize_grab(state: &mut State, button: u32) {
...
-            server::xdg_toplevel::ResizeEdge::BottomRight
-        }
-        _ => server::xdg_toplevel::ResizeEdge::None,
+    // If the window is floating, we use the floating_loc if available to prevent grabbing its shadow
+    let floating_loc = if window
+        .with_state(|state| state.layout_mode.is_floating() || state.layout_mode.is_floating())
+    {
+        window.with_state(|state| state.floating_loc())
+    } else {
+        None
```
Ottatop: The problem isn't whether the window is floating; it's that the `window_loc` above is the location of the pointer focus, which in this case is the shadow, not the window. This is an oversight on my part. We should instead get the location from the space with `state.pinnacle.space.element_location(&window)`.

src/api/window.rs
```diff
@@ -270,41 +270,49 @@ pub fn resize_grab(state: &mut State, button: u32) {
...
+        (bottom, server::xdg_toplevel::ResizeEdge::Bottom),
+        (left, server::xdg_toplevel::ResizeEdge::Left),
+        (right, server::xdg_toplevel::ResizeEdge::Right),
+    ];
+
+    let edges = borders
+        .into_iter()
+        .min_by_key(|(loc, _)| ((loc.x - pointer_loc.x).pow(2) + (loc.y - pointer_loc.y).pow(2)))
+        .map(|(_, edge)| edge)
+        .unwrap_or(server::xdg_toplevel::ResizeEdge::None);
```
Ottatop: Hm, using the nearest edge/corner point to determine the resize edge doesn't feel very intuitive. There are some points where I'd expect a corner resize but am getting an edge resize instead.

A few comments on the current logic:
1. Is the expression in `min_by_key` supposed to be Pythagorean distance? If so, it's missing the square root.
2. If it is Pythagorean distance, you can use `f64::hypot`.

A quick glance at Awesome's implemention shows that they divide the window into a 3x3 grid, where the outer tiles of the grid map to each corner/edge, and the logic for the center tile is the same as the logic prior to this PR. We should use that logic, I think it's more intuitive. If you have any suggestions for different logic I'm also open to that.

Ottatop (changes requested):

src/api/window.rs
Ph4ntomas: I had a first implementation that used 8 rectangles to have fixed regions, and 4 point w/euclidian distance, but scrapped it because this version was simpler (in term of code) iirc I've kept it. I'll recover it.

> Is the expression in min_by_key supposed to be Pythagorean distance? If so, it's missing the square root.

The square root is not needed because a² > b² implies a > b.
We don't really care about the actual distance, only their order

> If it is Pythagorean distance, you can use f64::hypot.

Oh, I did not know that function

src/api/window.rs
Ph4ntomas: Ok, I'll check that out !

src/api/window.rs
Ph4ntomas: Regarding the grid, I could either use a factor of 3 for each dimension, or something 'fancier' where the corner have a smaller size (e.g. make the corner squares of 1/6th the height or width, whichever is smallest).

Do you have a preference, or should I go with whatever feels right, then adjust later on ?

I personally think that 1/3 makes the corner too big, and rectangular area for corner feels weird, but it's not that big of an issue either

Ph4ntomas: > We should use that logic, I think it's more intuitive. If you have any suggestions for different logic I'm also open to that.

I went with a simple 3x3 grid with proportional size. However, falling back to corner resize on the center feels weird to me, so I've changed the logic to use edges instead.

I've tested both behavior. If you prefer the corner behavior for the center, let me know and I'll push it instead.

Ottatop: > I went with a simple 3x3 grid with proportional size.

Yea that was my preference.

> However, falling back to corner resize on the center feels weird to me, so I've changed the logic to use edges instead.
>
> I've tested both behavior. If you prefer the corner behavior for the center, let me know and I'll push it instead.

I've been giving this some thought: can we experiment with splitting the center tile into a 3x3 grid and re-running the logic on that tile, repeating the center splits until an edge is found? I think this could be a best-of-both-worlds solution.

Ph4ntomas: > can we experiment with splitting the center tile into a 3x3 grid and re-running the logic on that tile, repeating the center splits until an edge is found? I think this could be a best-of-both-worlds solution.

It's not hard to implement, but the behavior would not be 'logical', since the closer you move toward the center point, the smaller the areas.

This has two direct effects:
- close to the center, the behavior become essentially random from a user perspective
- even before that, missing the top edge lower bound by a few pixels downward put you into corner resize mode, but only sometime.

Since there are no visual indicators on the region you're about to click, I don't think it's a good idea.

Ph4ntomas: I'm off my computer right now, but I can implement the recursion tomorrow (UTC)

Ph4ntomas: On the other hand, I could extend the logic for the center to 8 directions, and either apply that on the center, or the whole window. (Imo it's better to do that only in the center, since it's more in line with other compositors behavior)

Ottatop: Hm yea, just did a quick and dirty impl of the continuous slicing and it's a little jank.

I think you've made some good points about having the center be edge-resize only, so let's go with that.

Ph4ntomas: Okay, then I think the PR should be good to go (feature-wise), let me know if you see anything else during the review

src/api/window.rs
```diff
@@ -269,42 +269,87 @@ pub fn resize_grab(state: &mut State, button: u32) {
...
+        (0, 1) => server::xdg_toplevel::ResizeEdge::Left,
+        (2, 1) => server::xdg_toplevel::ResizeEdge::Right,
+
         _ => server::xdg_toplevel::ResizeEdge::None,
     };
 
+    let edges = if edges != server::xdg_toplevel::ResizeEdge::None {
+        edges
+    } else {
+        // A bit of an explanation here.
```
Ottatop, line 310: Took me like 5 minutes to figure out what was going on here, think this needs a short explainer
```suggestion
        // Find the closest edge by figuring out which corners the pointer is between.
        // This works by drawing lines from the window's center to all four corners and the pointer.
        // Whichever two lines the pointer line is between determines the edge chosen.
        
        // A bit of an explanation here.
```

Ottatop (approved):
Lgtm, thank you!

Issue #324, opened 2025-06-30
Various new actions


Ottatop:
Some new actions for v0.2:

Definitely getting added:
- [x] `WindowHandle::output`
    - Returns the output a window is on/the output of the window's first tag
- [x] `WindowHandle::swap`
    - Swaps a window with another
    - *Might* allow you to swap two windows in different layout modes rather than just tiled windows, idk yet

Possibly getting added:
- [x] `OutputHandle::active_tags`
    - Like `OutputHandle::tags` but filters for active tags
- [x] `WindowHandle::move_to_output`
    - To move a window to another output by setting its tags to that output's active tags 
- [x] `WindowHandle::set_tags`
    - To set exact tags from a list of tags
    - Not to be confused with `WindowHandle::set_tag` without the s

Ph4ntomas:
Can we/I add a `lower` rpc to the WindowHandle ?

The way I see it, it would do the opposite of `raise` (so push the window to the bottom of the ZIndex stack. This would help for QoL when using mainly keyboard, to be able to recover window that are under the currently focused one without needing to cycle over every window to raise them individually.

Implementation wise, this would mean re-creating the ZIndex stack vector by draining the existing one into the new one (which would only contain the window). I think it's OK on a perf standpoint (I doubt anyone will have thousands of opened window)

Another nice-to-have RPC would be something to compute the (relative) ZIndex of a window, but that might be more niche

PR #337, opened 2025-07-13
api/window: Add a RPC to move a window to other output

### LINKED ISSUE #160: Enable inter-output window management
Currently, everything related to window management is limited to within any given output. Dragging windows to other outputs will not update their tags, and there's nothing in the API to move windows to other outputs.

Goals:
- [ ] Make dragging windows to other outputs update tags
- [ ] Add moving windows to other outputs to the API

Ph4ntomas:
This PR adds a new RPC to move a window to a given output.

This is done by setting the window tags to the output ones, computing a
new floating location (to prevent the window jumping back), and
refreshing the layout of the original and new output for spilled window.

It should closes #160, and update #324


tests/integration/api/window.rs
```diff
@@ -1424,6 +1427,58 @@ fn window_handle_in_direction() {
...
+        let client_id = fixture.add_client();
+
+        let surface = fixture.spawn_floating_window_with(client_id, (500, 500), |_| ());
+
+        fixture
+            .client(client_id)
+            .window_for_surface(&surface)
+            .ack_and_commit();
+        fixture.roundtrip(client_id);
+
```
Ph4ntomas: @Ottatop I' m not sure I understand why this is needed. 

I've added this because there is a pending transaction at that point. Without it, the configure changes are 'lost' when we ack the configure later one.

I believe it's due to the pending transaction/state/configure not being invalidated, which it should if the client hasn' t ack it at that point.

Ph4ntomas (commented):
@Ottatop This is overall working, but there is some weird behavior that I'm not sure how to fix.

The first issue I see is that `kitty` draw its window decoration outside the window area if the window was floating initially, made fullscreen, moved and then un-fullscreened. This effect last until the window lose focus, so I'm not really sure what the cause is. To me, it looks like some issue with the configure/pending state in `handlers/window.rs`.

In the test, I noticed that `ack_and_commit` wasn't working if there was a pending_transaction before calling the method. I believe this is linked to the aforementioned issue, but I'm not sure how to troubleshoot this further.

Do you have an idea somewhere I should look to ?

Ottatop:
Travelling until Thursday and don't have a second monitor, I'll look into it when I get back unless I can figure it out beforehand

Ph4ntomas:
> The first issue I see is that kitty draw its window decoration outside the window area if the window was floating initially, made fullscreen, moved and then un-fullscreened. This effect last until the window lose focus, so I'm not really sure what the cause is. To me, it looks like some issue with the configure/pending state in handlers/window.rs.

This is a kitty bug. When they receive a configure event, they re-compute the viewport size, which takes into account the CSD, so both the CSD and kitty's viewport fit in the configured area. The issue is that when leaving fullscreen mode, this computation is done before the CSD surface is created, leading to the viewport taking the whole space, and the CSD being drawn above it.

I'll open a bug report on kitty's github.

tests/integration/api/window.rs
Ph4ntomas: Okay, I've found the issue.

There are two cases when a window position gets updated:
1. If a configure has been sent (either the window size changed, or there were change in its `xdg_toplevel::state`), pinnacle add a `MapTo()` transaction, and wait for the client to ack the configure instead of moving the window right away
2. If no configure is being sent, the window is mapped right away.

In this test case, the first ack_and_commit prevent the following: 
1. the client has a pending configure, tied to the initial window placement.
2. We then call move_to_output, which only changes the window location and doesn't trigger a new configure, so the window is mapped instantly.
3. The client ack the configure and pinnacle map it at the tied `MapTo` location.

This behavior breaks the transaction logic.

@Ottatop In my opinion this should be handled in another PR, since it's not really a trivial change.

Ph4ntomas: > In my opinion this should be handled in another PR, since it's not really a trivial change.

The last two commit fixes the transaction logic for this specific case. I've included them here because is was less complex that I thought.

There are other unconditional `map` that may need to be tied to pending locations:
src/grab/move_grab.rs:144 -> called while moving the window. The move itself should never cause a transaction to be generated, but one could be generated by the config. Changing this may prevent the window jittering if that happen (due to it's location being updated for a frame, and moved again the next frame).
src/output.rs:272 -> Same as above. This is called when the output scale changes.
src/handlers/window.rs:75 -> called if there was nothing to configure && no pending transaction. This one is safe, but I'm not sure if it's really useful.
src/handlers/xwayland.rs:124 & src/handlers/xwayland.rs:124 -> Not sure about XWayland related stuff.

I haven't squash the tests so I can split them off in another PR if needed, it was just convenient not to change branches for the time being.

Ottatop: > The last two commit fixes the transaction logic for this specific case. I've included them here because is was less complex that I thought.

Think it's fine to keep them in this PR if it's that simple

> There are other unconditional `map` that may need to be tied to pending locations:

I suppose we could do this in another PR to centralize the mapping down to one spot. The xwayland stuff is for override-redirect windows that need to be mapped at a specific point (they're like Wayland popups but it's windows because xwayland 🫠).

PR #339, opened 2025-07-19
api: add [in]active tags filtering to OutputHandle

Ph4ntomas:
This PR adds part of #324 to the API, namely the OutputHandle::active_tags method.

Since protobuf allows to add new field in a backward compatible way, I've just added a filter argument.
The filter argument is setup to fallback on the 'All' filter, so client/config built for the previous API will still behave as expected.

The client/config facing api gets an additional `active_tag` method, and I've added an inactive method as well.

### DIFF
--- a/api/lua/pinnacle/output.lua
+++ b/api/lua/pinnacle/output.lua
@@ -756,6 +756,62 @@ function OutputHandle:tags()
     return handles
 end
 
+---Gets the active tags this output has.
+---
+---@return pinnacle.tag.TagHandle[]
+function OutputHandle:active_tags()
+    local tags = self:tags()
+
+    ---@type (fun(): boolean)[]
+    local batch = {}
+    for i, tag in ipairs(tags) do
+        batch[i] = function()
+            return tag:active()
+        end
+    end
+
+    local responses = require("pinnacle.util").batch(batch)
+
+    ---@type pinnacle.tag.TagHandle[]
+    local active_tags = {}
+
+    for i, is_active in ipairs(responses) do
+        if is_active then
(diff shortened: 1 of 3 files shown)


api/protobuf/pinnacle/output/v1/output.proto
```diff
@@ -130,8 +130,15 @@ message GetFocusedResponse {
   bool focused = 1;
 }
 
+enum TagFilter {
+  ALL = 0;
+  ACTIVE = 1;
+  INACTIVE = 2;
+}
```
Ottatop: For future work, enum value names should follow the [protobuf style guide](https://protobuf.dev/programming-guides/style/#enums) like the other enums.

Ottatop (commented):
I think instead of extending the `GetTagIdsRequest` with a filtering enum, we should instead do something similar to [`OutputHandle::keyboard_focus_stack_visible`](https://github.com/pinnacle-comp/pinnacle/blob/54ab3ca1549898c40b50737e5edfe478ca7c4d66/api/rust/src/output.rs#L935-L941) and filter for tag active-ness client-side. Yes, it's a bunch more API calls, but I'm not sure I want to commit to putting filtering directly in the protobuf (idk how many people are going to be using `inactive_tags()` for example). We can always revisit performance down the line.

api/protobuf/pinnacle/output/v1/output.proto
Ph4ntomas: Got it, that make sense. For some reason I thought protobuf enums were scoped

Ottatop (approved):
Looks good, thanks!

PR #340, opened 2025-07-19
window: Add method to swap two windows

Ph4ntomas:
Adds `WindowHandle::swap`, as mentioned in #324.

In my opinion, this method should allow swapping with *any* windows, as log as said window has passed the initial configure (i.e. as long as the window is swapped).

The initial commit only allows swapping tiled window on a single output, since that remove a lot of complexity for an initial implementation.

Todo:
- [x] Initial implementation, only allows swapping two tiled window in the same output.
- [ ] Allow swapping regardless of the layout mode.
- [ ] Allow swap to work cross output.


Ph4ntomas:
Some of the (existing) test cases are inconsistent because it's possible to leave the scope where api calls are made before the compositor is done. 
This leads to tests randomly failing or getting in an infinite loop.

I've somewhat mitigated the issue in the tests I've added (mostly busy waiting for configure serials to change, and/or some identifiable states to be reached before proceeding). 
I'm trying to think of a cleaner/more generic way to do so, and I'll refactor the existing at well when I find it.

api/protobuf/pinnacle/window/v1/window.proto
```diff
@@ -213,6 +218,7 @@ service WindowService {
   rpc Raise(RaiseRequest) returns (google.protobuf.Empty);
   rpc MoveGrab(MoveGrabRequest) returns (google.protobuf.Empty);
   rpc ResizeGrab(ResizeGrabRequest) returns (google.protobuf.Empty);
+  rpc Swap(SwapRequest) returns (google.protobuf.Empty);
```
Ottatop: Can we have this return an empty `SwapResponse` message? I've been trying to have all new rpcs return a message that we control so we can extend them if necessary, there was an instance where I had to change `google.protobuf.Empty` to a custom message a while back.

api/rust/src/window.rs
```diff
@@ -817,6 +817,21 @@ impl WindowHandle {
         response.window_ids.into_iter().map(WindowHandle::from_id)
     }
 
+    /// Swap position with another window.
+    pub fn swap(&self, target: WindowHandle) {
```
```suggestion
    pub fn swap(&self, target: &WindowHandle) {
```
This method doesn't need to consume the handle.

api/rust/src/window.rs
```diff
@@ -817,6 +817,21 @@ impl WindowHandle {
...
+        self.swap_async(target).block_on_tokio()
+    }
+
+    /// Async impl for [`Self::swap`].
+    pub async fn swap_async(&self, target: WindowHandle) {
```
```suggestion
    pub async fn swap_async(&self, target: &WindowHandle) {
```
Same here

api/lua/pinnacle/window.lua
```diff
@@ -743,6 +743,22 @@ function WindowHandle:in_direction(direction)
     return response and window_handle.new_from_table(response.window_ids or {}) or {}
 end
 
+---Swap position with another window.
+---
+---@param target pinnacle.window.WindowHandle
+function WindowHandle:swap(target)
+    if target == nil or target.id == nil then
+        log.error("Invalid window handle")
```
Ottatop, line 751: Probably should return here

src/api/window.rs
```diff
@@ -366,3 +366,70 @@ pub fn resize_grab(state: &mut State, button: u32) {
...
+        state.floating_x = window_floating_loc.map(|loc| loc.x);
+        state.floating_y = window_floating_loc.map(|loc| loc.y);
+        state.floating_size = window_floating_size
+    });
+
+    window.with_state_mut(|state| {
+        state.floating_x = target_floating_loc.map(|loc| loc.x);
+        state.floating_y = target_floating_loc.map(|loc| loc.y);
+        state.floating_size = target_floating_size
+    });
```
Ottatop: We need to hold `floating_x` and `floating_y` separately i.e. `let (window_floating_x, window_floating_y) = window.with_state(|state| (state.floating_x, state.floating_y));`, using `state.floating_loc()` may lose one of the axes if one is None and the other isn't.

Though to be fair there's some other places where we just straight up lose the floating location so I guess this change is optional until I figure out a better system for tracking locations

api/protobuf/pinnacle/window/v1/window.proto
Ph4ntomas: Ok, make sense. 

I guess this should be done for the MoveToOutput method too before 0.2.0 land ? Do you want me to open a PR for that ?
 (I´m assuming that kind of change is ok between release. If not, well I'll keep that in mind for future PR)

src/api/window.rs
Ph4ntomas: > using state.floating_loc() may lose one of the axes if one is None and the other isn't.

It's unclear to me why this would ever happen, so I did not think about it. Out of curiosity, do you have an example ?

Ottatop: If you go to the implementation of `floating_loc()`, it returns Some only when both `floating_x` and `floating_y` are Some. If one of those is none, the current swap implementation loses the other one.

api/protobuf/pinnacle/window/v1/window.proto
Ottatop: Ah, missed that. Yes, PR sounds good. All breaking changes are ok in the development period between releases.

src/api/window.rs
Ph4ntomas: Yes, but I could not find a reason for either to be None in the codebase, while the other is set.

I have changed the code btw, I'm just wondering how we could have the floating location only partially set, or what that would represent

Ottatop: All fields in `WindowHandle::set_geometry` are optional so you can set an x but not a y, [here](https://github.com/pinnacle-comp/pinnacle/blob/7d7da249c2d72130d00ce1bfddcbde5a34c0f57a/src/api/window.rs#L32-L63). I had to split the location out for scenarios like setting the x but not the y when a window is unmapped, because there's no good y to default to in that case.

Ph4ntomas: Oh, I see, make sense then.

PR #341, opened 2025-07-21
api/window: add response to WindowHandle::MoveToOutput RPC

Ph4ntomas:
As mentioned in #340, this adds a typed response to window::MoveToOutput RPC


Ottatop (approved):
Thank you!

Issue #144, opened 2024-01-24
Add pointer and keyboard controls to API


Ottatop:
This includes:
- Ability to click and move the mouse from a config
- Ability to send press and release events for keys from a config
- Getters for the current pointer position

PR #342, opened 2025-07-21
Add server-side decorations

Ottatop:
This PR adds server-side decorations drawn with Snowcap.

Todos:
- [x] Set the decoration protocol in stone
- [x] Get the config decoration API up and running
- [x] Support multiple decoration surfaces per window
- [x] Decoration pointer input
- [x] Decoration input regions
- ~~Figure out if or how keyboard focus to decoration surfaces is gonna work~~
- [x] Take into account decoration bounds when calculating tile geometries
- [x] Take into account decoration bounds when using `WindowHandle::set_geometry` and getters for loc and size
- [x] Synchronize decoration surface resizing with window resizing
    - Currently this is window resizes -> send a configure to decoration surface with new size -> decoration surface resizes. We'll need to block updating the window until the decoration surface finishes resizing.
    - This will add noticeable latency just from looking at how resizing a floating window affects the decoration surface currently. May need to think of a better way to do this.
    - Update: most of the slowness is from Snowcap being slow to resize everything. Maybe look into multithreading widget rendering.
    - Update update: Ok this might be quite hard to solve. With vulkan, resizing is slow because there are a bunch of Wayland roundtrips when configuring surfaces. Multithreading this is not trivial because of vulkan validation errors. The roundtrip problem doesn't happen with tiny-skia software rendering, but tiny-skia is quite slow to render, *especially* with text, which is bad because the bind overlay has a lot of it.
    - Update update update: Compromise: we're now just forcing tiny-skia rendering for decorations, layer widgets render with vulkan.
- [x] Hook protocol/config signals into the decoration API to let it update for things like window focus changing
    - Currently this is just letting users send messages to the decoration to get it to update 
- [x] Support decoration surface location changes (have only tested with x=y=0 so far)
- [x] Test building without the `snowcap` feature
- ~~Handle decoration unmapping~~ Yea we're just gonna not have this
- [x] Add way to set bounds and other properties dynamically
- [x] Include decoration surfaces for window pointer focus
- [x] Ignore decorations when fullscreen
- [x] Code cleanup and dedup
- ~~Perf testing~~ Moving off to a big perf evaluation before 0.2
- [x] Believe there's a panic on exit when a decoration surface exists somewhere? Investigate that
- [x] Probably a bunch of other things

A future PR after this:
- Built-in focus border
- Built-in window titlebars


Ph4ntomas:
> Figure out if or how keyboard focus to decoration surfaces is gonna work

Just my two cents, but I don't think we need keyboard focus with decorations surface.

When I'm using keyboard only interaction, I already have keybindings for most actions window decorations offer (maximize/fullscreen/minimize/close etc.). 

The only case where that might be needed is a binding to open&focus menus, or re-focus said menus if they are part of the decoration surface, but I don't think the current protocol allows client to expose a menu in SSD anyway.

As a side note, #144 would also allow interracting with SSD from the keyboard, so again, I'm not sure there is a real need for keyboard focus

Ottatop:
I was thinking of more esoteric use cases like someone putting a text input widget on a SSD, but I suppose we can kick that can down the road.

> As a side note, https://github.com/pinnacle-comp/pinnacle/issues/144 would also allow interracting with SSD from the keyboard, so again, I'm not sure there is a real need for keyboard focus

This is more keyboard/mouse emulation. A SSD would still need to be able to gain keyboard focus for a key press to be sent to it.

Ph4ntomas:
> I was thinking of more esoteric use cases like someone putting a text input widget on a SSD, but I suppose we can kick that can down the road.

Ah, yeah, I did not think of that use-case

> > As a side note, https://github.com/pinnacle-comp/pinnacle/issues/144 would also allow interracting with SSD from the keyboard, so again, I'm not sure there is a real need for keyboard focus
>
> This is more keyboard/mouse emulation. A SSD would still need to be able to gain keyboard focus for a key press to be sent to it.

My point was that it would allow to emulate any mouse interaction lacking kbd binding without needing actual keyboard focus on the SSD button (but now that you've mentioned the 'text input in SSD' usecase, it doesn't change anything for that use-case)

PR #344, opened 2025-07-22
Add WindowHandle::lower

Ph4ntomas:
This PR adds a RPC to lower a window to the back.

I mentioned it in #324, when the main means of interacting with the compositor is through keyboard inputs, it's a bit tedious to switch to a lower window, since you have to basically guess where the window is. This gives tools to lower the currently focused window to the back of the Z-Buffer.

One thing I'm unsure of is whether lowering a fullscreen/maximized window make sense.
I've allowed it because in my opinion, RPCs is in the API should gives building block for configuration, not restrict whats possible as long as it does not leave the compositor in a bad state.

### DIFF
--- a/api/lua/pinnacle/window.lua
+++ b/api/lua/pinnacle/window.lua
@@ -562,6 +562,17 @@ function WindowHandle:raise()
     end
 end
 
+---Lowers a window.
+---
+---This will bring the window to the back.
+function WindowHandle:lower()
+    local _, err = client:pinnacle_window_v1_WindowService_Lower({ window_id = self.id })
+
+    if err then
+        log.error(err)
+    end
+end
+
 ---Returns whether or not this window is on an active tag.
 ---
 ---@return boolean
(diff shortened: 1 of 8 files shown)


api/lua/pinnacle/window.lua
```diff
@@ -562,6 +562,17 @@ function WindowHandle:raise()
     end
 end
 
+---Lower a window.
```
Ottatop:
```suggestion
---Lowers a window.
```

src/focus.rs
```diff
@@ -209,6 +209,23 @@ impl Pinnacle {
         self.update_xwayland_stacking_order();
     }
 
+    /// Lower a window to the bottom of the z-index stack.
+    pub fn lower_window(&mut self, window: WindowElement) {
+        let _span = tracy_client::span!("Pinnacle::raise_window");
```
Ottatop:
```suggestion
        let _span = tracy_client::span!("Pinnacle::lower_window");
```

api/rust/src/window.rs
```diff
@@ -577,6 +577,15 @@ impl WindowHandle {
             .unwrap();
     }
 
+    /// Lower this window to the back.
```
```suggestion
    /// Lowers this window to the back.
```

Ottatop (approved):
Thanks!

PR #345, opened 2025-07-22
Decrease tests flakyness

### LINKED ISSUE #343: Window related tests are failing randomly
### Describe the bug

Some tests (mostly those requiring the client to acknowledge a configure) can fail at random.

I observed to kind of failure:
- Test failing
- Test locking-up/being in an infinite loop.

### Reproduction steps

Run tests a bunch of time

### Relevant parts of config

N/A

### Version info

pinnacle 0.1.0

Build Info:
    Build profile: debug
    rustc version: 1.88.0
    Branch: main
    Commit: 4efe15f (Merge pull request #341 from Ph4ntomas/window_handle-move-to-output_fix-return)

### Distro

Arch

### GPU

Irrelevant

### Config language

Lua

Ph4ntomas:
This is an attempt to reduce test flakyness by waiting for some events before proceeding.

Namely:
- wait for pending transactions to be empty
- waiting for the client to receive a configure event when we expect one.

As far as I can tell, this fixes most spurious failure caused by some assertion running before the state was updated.

What isn't fixed (yet) is the event loop locking up in some tests, ~~which make some tests wait indefinitely.~~
EDIT: No more infinite loop. The `dispatch_until` function was causing these, and can now timeout. This makes the tests fail, but at least it fails with output & backtrace, and it's not half the tests that are stuck.

Fixes #343

### DIFF
--- a/api/lua/pinnacle/tag.lua
+++ b/api/lua/pinnacle/tag.lua
@@ -93,6 +93,7 @@ function tag.get(name, output)
     output = output or require("pinnacle.output").get_focused()
 
     if not output then
+        log.error("Cannot get tags without output")
         return
     end
 
@@ -334,6 +335,10 @@ end
 function TagHandle:name()
     local response, err = client:pinnacle_tag_v1_TagService_GetName({ tag_id = self.id })
 
+    if err then
+        log.error(err)
+    end
+
     return response and response.name or ""
 end
 
@@ -342,6 +347,10 @@ end
 ---@return pinnacle.output.OutputHandle
 function TagHandle:output()
     local response, err = client:pinnacle_tag_v1_TagService_GetOutputName({ tag_id = self.id })
+
+    if err then
+        log.error(err)
+    end
 
     local output_name = response and response.output_name or ""
     return require("pinnacle.output").handle.new(output_name)


Ph4ntomas:
Hopefully we shouldn't see timeouts now.

I'm leaving the timeout behavior since it's good to know if a test is badly behaved, but hopefully if we trip it from now on, it will be for a legitimate reason.

The last few commits ([c6856a0](https://github.com/pinnacle-comp/pinnacle/pull/345/commits/c6856a03303e1943f2ab4698529276efe16c21ca) and after) are some changes to the logging I did to help me troubleshoot the tests. Since I found them useful, I've added them here. 

If heavy changes are needed on these commits, let me know, I'll move them on another PR to unblock this one.
Same if you me want to drop them (although I think we should at least keep c6856a0 and the `__tostring` part of 00c0cf5, but I can open a new PR if needed)

api/lua/pinnacle/grpc/stream_extension.lua
Ottatop, line 1: The contents of this file instead need to be copy-pasted into `api/lua/build/grpc_client.lua` so it can get generated separately for both the pinnacle and snowcap apis.

tests/integration/common/fixture.rs
```diff
@@ -78,9 +80,14 @@ impl Fixture {
             event_loop,
             state,
             _test_guard,
+            timeout: Duration::from_secs(DEFAULT_TIMEOUT),
         }
     }
 
+    pub fn _set_timeout(&mut self, timeout: Duration) {
+        self.timeout = timeout;
+    }
```
Ottatop: Is this used anywhere?

tests/integration/common/fixture.rs
```diff
@@ -35,6 +36,7 @@ struct State {
 }
 
 static OUTPUT_COUNTER: AtomicU32 = AtomicU32::new(0);
+static DEFAULT_TIMEOUT: u64 = 60;
```
Ottatop: This should be a const `Duration` instead of a static `u64`

tests/integration/common/fixture.rs
```diff
@@ -257,19 +296,59 @@ impl Fixture {
...
     }
 
     pub fn client(&mut self, id: ClientId) -> &mut Client {
         self.state.client(id)
     }
+
+    /// Wait for all outstanding transaction and configure to be handled.
+    ///
+    /// If timeout is reached, stop with false.
+    pub fn flush(&mut self, timeout: Duration) -> bool {
```
Ottatop: If all the calls to flush have an `assert` around them, I think we should just panic on timeout here instead of returning a bool.

tests/integration/common/fixture.rs
```diff
@@ -257,19 +296,59 @@ impl Fixture {
...
+            if start.elapsed() >= timeout {
+                return false;
+            }
+        }
+
+        true
+    }
+
+    /// Wait for a client to receive a configure, dispatching in the meantime.
+    pub fn wait_client_configure(&mut self, id: ClientId, timeout: Duration) -> bool {
```
Ottatop: Same here

tests/integration/api/window.rs
```diff
@@ -145,11 +147,8 @@ fn window_handle_set_geometry_floating() {
             },
         }
 
-        fixture
-            .client(client_id)
-            .window_for_surface(&surface)
-            .ack_and_commit();
-        fixture.roundtrip(client_id);
+        assert!(fixture.wait_client_configure(client_id, Duration::from_secs(10)));
+        assert!(fixture.flush(Duration::from_secs(10)));
```
Ottatop: We should make a const duration and use it in place of all these `from_secs` so we can change it easily if needed. Or maybe even just remove the timeout param altogether and use the fixture's `timeout` field.

tests/integration/api/window.rs
```diff
@@ -183,11 +182,7 @@ fn window_handle_set_geometry_tiled_does_not_change_geometry() {
             },
         }
 
-        fixture
-            .client(client_id)
-            .window_for_surface(&surface)
-            .ack_and_commit();
-        fixture.roundtrip(client_id);
+        assert!(fixture.flush(Duration::from_secs(10)));
```
Ottatop: Is there supposed to be a `wait_client_configure` before this?

Ottatop (commented):
Think it's fine to keep the logging stuff in this PR.

tests/integration/api/window.rs
Ph4ntomas: Yeah agreed.

I don't think I want to use the fixture timeout here, as I'd rather keep that one larger.

10s is already way overkill on these, but indeed , a const or a second 'wait pending timeout' on the fixture would be better a good idea

tests/integration/common/fixture.rs
Ph4ntomas: Not anymore. 

I'll remove the method for the time being, but I'll leave in the member variable. I'd rather keep the timeout configurable if we need to increase it for some tests at some point.

tests/integration/api/window.rs
Ph4ntomas: `wait_client_configure` will block if the client never receive a configure.
Here tiled clients should not receive a configure, so the function will timeout (and panic with the requested changes).

`flush` is there to wait for transaction if one is generated and prevent the test from succeeding due to a race-condition. The function will ack configure as needed to make the transactions proceed.

tests/integration/api/window.rs
Ph4ntomas: > I don't think I want to use the fixture timeout here, as I'd rather keep that one larger.

I shouldn't have commented before grabbing a coffee. Since the test fails either way, there's no point in over-engineering this. I'll use the fixture timeout.

PR #348, opened 2025-07-31
Refactor window mapping

Ph4ntomas:
This PR is an effort to refactor and unify how and when window are mapped onto screen (understand moved), as well as what output are scheduled to render when this happen.

Here is the current state:
- XWayland handlers map directly to space, but:
  - `mapped_override_redirect_window` additionally and schedule outputs to render,
  - `configure_notify` map the doesn´t schedule outputs to render,
- Changing a window layout may map directly, and what outputs are scheduled to render depends on what behavior was used.

TODO:
- [x] Always go through a transaction when setting a window location
- [ ] Only schedule output to refresh when mapping the window (this allow to refresh output only once)
- [ ] pick output before and after the mapping occur so we also schedule the output the window was previously on.
- [ ] decide whether refactoring `update_layout_from_tree` and `update_layout_mode_and_layout` functions make sense
- [ ] refactor window swapping such that they use a single translation if that make sense.


src/window/layout.rs
Ph4ntomas, line 1: The more I think about this file, the less I'm sure about it being here, or being named that way. This is a mix of functions spawned from refactoring src/layout.rs and src/handler/windows.rs and is window & layout related, so it kinda fit.

On the other hand it very much feels like functions are there because I wasn't sure where to put them (which is kinda the case tbh).

Unmap_window could go to layout.rs, but that file is already crowded. However this is very much the odd one.

The `*map*` functions could go with utils/transaction.rs, but I feel it would be weird.

The update_layout && update_geometry are still doing a bit too much. On the other hand I don't know if it's really clever to remove the mapping part from them as it stand, sice we'd need a way to return the location at the very least, or store it in window element, and have a way to know whether we need to map the window, or add them to the pending_window list and trigger a new layout.

Tbh it feels like having Pinnacle impls block split into nearly every file make it unclear where it would be obvious to find these functions. That being said it's a side effet of the pinnacle object doing almost everything, but I don't feel confident trying to refactor that at the moment.

Anyway, this is to say I'm open to moving suggestion as to where to move the functions around (or how to refactor further if needed)

src/api/window.rs
```diff
@@ -421,23 +426,55 @@ pub fn swap(state: &mut State, window: WindowElement, target: WindowElement) {
...
+    let window_geo = state.pinnacle.space.element_geometry(&window);
+
+    let target_layout_mode = target.with_state(|state| state.layout_mode);
+    let target_geo = state.pinnacle.space.element_geometry(&target);
+
+    let mut builder = TransactionBuilder::new();
+    let mut unmappings = Vec::new();
+
+    if window.is_on_active_tag() {
+        let geo = target_geo.expect("Target should have had a geometry");
```
Ottatop: Can we extract `window.is_on_active_tag()` to before the actual tag swap, something like `let target_was_on_active_tag = target.is_on_active_tag();`? Right now it reads like it's expecting the wrong geometry.

src/api/window.rs
```diff
@@ -421,23 +426,55 @@ pub fn swap(state: &mut State, window: WindowElement, target: WindowElement) {
...
-        state.update_window_layout_mode_and_layout(&target, |layout| {
-            layout.apply_mode(window_layout_mode)
-        });
-    } else {
-        state.pinnacle.request_layout(&output);
-        if output != target_output {
-            state.pinnacle.request_layout(&target_output);
-        }
+    if target.is_on_active_tag() {
+        let geo = window_geo.expect("Window should have had a geometry");
```
Ottatop: Similar thing here

src/api/window.rs
```diff
@@ -421,23 +426,55 @@ pub fn swap(state: &mut State, window: WindowElement, target: WindowElement) {
...
+        state
+            .pinnacle
+            .unmap_window(&mut state.backend, &target, &target_output)
+    {
+        target.with_state_mut(|state| state.layout_mode.apply_mode(window_layout_mode));
+        unmappings.push(unmapping);
     }
+
+    // We need one output here. I've picked the one the window was on, although I doubt it matters
+    // in this specific case. I guess the alternative would be one TB per output.
```
Ottatop, line 474: Think this is fine

wlcs_pinnacle/src/main_loop.rs
```diff
@@ -155,6 +155,7 @@ fn handle_event(event: WlcsEvent, state: &mut State) {
                     // state.window_state.set_floating(true);
                 });
 
+                // I'm not sure about this one.
```
Ottatop: We can assume wlcs windows are always floating, I think it's fine to keep this as a simple map.

src/window/window_state.rs
```diff
@@ -486,6 +489,45 @@ impl WindowElement {
             }
         }
     }
+
+    /// Apply current mode layout mode to the window underlying surface
+    ///
+    /// Toplevel need a call to `send_configure` or `send_pending_configure` for these changes to
+    /// be effectives.
+    pub(crate) fn configure_states(&self) {
+        let _span = tracy_client::span!("Window::configure_states");
```
Ottatop:
```suggestion
        let _span = tracy_client::span!("WindowElement::configure_states");
```

src/window/layout.rs
```diff
@@ -0,0 +1,279 @@
...
+        let _span = tracy_client::span!("Pinnacle::update_window_layout_mode");
+        let output = window.output(self);
+
+        let Some(output) = output else {
+            tracing::error!("Cannot update the state of a window with no output");
+            window.configure();
+            return;
+        };
+
+        let Some(_) = self.space.output_geometry(&output) else {
```
Ottatop:
```suggestion
        if self.space.output_geometry(&output).is_none() {
```

src/window/layout.rs
```diff
@@ -0,0 +1,279 @@
...
+                &output,
+                transaction_builder.into_pending(Vec::new(), self.layout_state.pending_swap, false),
+            );
+        }
+
+        // not sure if something should be done in case there's no output. Does it make sense to
+        // create an outputless transaction ?
+        // xwayland handlers did not check for the output before unconditionally mapping the
+        // window. Should that behavior be kept ?
+        // Other method had the output checked ahead of time, so no issue there.
```
Ottatop: If the window has no output then it has no tags, so it should be fine to do nothing here.

src/window/layout.rs
Ottatop: I think we should hold off on moving around other stuff until we move into an alpha or release candidate or something to let things settle. I'll need some time to figure out where I want these functions to be.

Ph4ntomas: I fully agree. 

Do you want me to leave it there or to move it back to `src/handler/window.rs` ? Tbh I moved some of the logic while refactoring, but most of that was then scrapped. I moved the remaining function because the logic was essentially in two place and it wasn't really making sense, but if you'd rather have the code in the old location for the time being I can move everything back.

src/api/window.rs
Ph4ntomas: Yeah, make sense. I'm going with `[window|target]_was_on_active_tag` .

src/window/layout.rs
Ottatop: Guess we can just leave it here for now

src/api/window.rs
Ph4ntomas: Nice. 

If it's ok with you, I'm leaving the comment in for future reference.

src/window/layout.rs
Ph4ntomas: Ok. I've removed the comment since this is clear :)

PR #349, opened 2025-08-03
XDG Shell protocol: Add tests to ensure protocol correctness 

Ph4ntomas:
This started as part of #348, because I noticed `set_maximize` might not be up to spec because of `update_window_layout_and_layout`.
Specifically: 
> If the surface was already maximized, the compositor will still emit a configure event with the "maximized" state.

and

>If the surface is in a fullscreen state, this request has no direct effect. It may alter the state the surface is returned to when unmaximized unless overridden by the compositor.

are not implemented.

I'm making it a standalone PR because I want to cover ~~the whole protocol~~ toplevel related functions, and it's possible this one will be ready before #348 (another possibility is that I split out refactoring `update_window_layout_and_layout` from #348 since that PR primary goal is already implemented).

### DIFF
--- a/src/window.rs
+++ b/src/window.rs
@@ -543,9 +543,19 @@ impl State {
             let Unmapped {
                 window,
                 activation_token_data: _, // TODO:
-                state: _,
+                state,
             } = unmapped;
-            (window, true, true)
+
+            // if it *was* configured, we still want to respect that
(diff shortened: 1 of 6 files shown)


Ottatop (approved):
Thank you!

PR #352, opened 2025-08-18
feat(nix): add packages and modules

cassandracomar:
## Description

this PR:
1. add a nix package that builds the main server binary.
2. add a function that only needs a few parameters to build a rust config binary.
3. add a nixos module that creates a wayland session, launchable from a display manager.
4. add a home-manager module to create the user target and write a configurable `pinnacle.toml` to `$XDG_CONFIG_HOME/pinnacle`.

all of these can be upstreamed to home-manager/nixpkgs once things stabilize a bit and a user could conceivably use a specific version rather than git HEAD (critical features are still coming a bit fast right now).

I've mostly tested with the [UWSM](https://github.com/Vladimir-csp/uwsm) approach as it fits better with how upstream nixos modules work. however, the user service should also work fine. leaning into the former across the board would allow distros to manage the session however they prefer so removing `pinnacle-session` is something to think about.

as a side note, I don't know that much about lua so I haven't tested using these with a lua config. would appreciate some help testing that -- I can provide assistance.

### Usage

first, we need to set up nixos -- if you're on a non-nixos distro, you'll need to handle this part however your distro handles these things:

1. add the pinnacle overlay when importing nixpkgs:
    ```nix
    pkgs = import inputs.nixpkgs {
      inherit system;
      overlays = [
        inputs.pinnacle.overlays.default
        # your other overlays
      ];
    }
    ```
1. import the nixos module from the flake -- it looks like this if you're using a flake based config and the `nixosSystem` function:
    ```nix
      nixpkgs.lib.nixosSystem {
        inherit pkgs;
        modules = [
          inputs.pinnacle.nixosModules.default
          # your other configuration modules
        ];
      }
    ```
1. set up pinnacle as a wayland session:
    ```nix
      programs.pinnacle = {
        enable = true;
        xdg-portals.enable = true;
        withUWSM = true;
      };
      # make sure to set the following if you're using auto-login
      services.displayManager = {
        defaultSession = "pinnacle-uwsm";
      };
  
      # not sure if the following is strictly necessary but you probably
      # want to make sure they're set somewhere:
      services.xserver.enable = true;
      programs.dconf.enable = true;
      security.polkit.enable = true;
      hardware.graphics.enable = true;
      # you should also make sure the xkb layout is configured
    ```

then we need to configure your users to launch pinnacle properly:

1. load the home-manager module:
    ```nix
    {
      home-manager.sharedModules = [inputs.pinnacle.hmModules.default];
    }
    ```
1. set up the user service/targets in one of your home-manager user modules:
    ```nix
      wayland.windowManager.pinnacle = {
        enable = true;
        # if you're using rust -- if you're using lua, I /think/
        # you can just write the script with `pkgs.writeScriptBin`
        # and provide that output but I'm not positive.
        clientPackage = pkgs.pinnacle.buildRustConfig {
          name = "pinnacle-config";
          src = path/to/pinnacle/rust/config;
          version = "0.1.0";
        };
        systemd = {
          enable = true;
          # use UWSM instead
          useService = false;
          xdgAutostart = true;
        };
      };
      # make sure to start a bar and a launcher of some kind
      # I've tested waybar and walker personally -- they work fine.
    ```

once we're ready to upstream, we need to get the nixos packages/modules upstream first, and once done we can submit a PR to home-manager.

### Addendum
`buildRustPackage` and `buildRustCrate` don't really handle workspace dependencies all that well -- or at least, I couldn't find decent documentation on how to make them handle workspaces properly. consequently, I had to modify `build.rs` for the api defs packages to allow passing an absolute path to the protobuf definition directories as the relative paths are wrong when using `buildRustPackage`.

### DIFF
--- a/wiki/getting-started/installation.md
+++ b/wiki/getting-started/installation.md
@@ -19,6 +19,81 @@ yay -S pinnacle-comp-git
 
 :::
 
+### NixOS
+First, we need to set up Pinnacle as a session loadable by your display manager:
+
+1. add the pinnacle overlay when importing nixpkgs:
+    ```nix
+    pkgs = import inputs.nixpkgs {
+      inherit system;
+      overlays = [
+        inputs.pinnacle.overlays.default
+        # your other overlays
+      ];
+    };
+    ```
+1. import the nixos module from the flake -- it looks like this if you're using a flake based config and the `nixosSystem` function:
+    ```nix
+      nixpkgs.lib.nixosSystem {
+        inherit pkgs;
+        modules = [
+          inputs.pinnacle.nixosModules.default
+          # your other configuration modules
+        ];
+      }
+    ```
+1. set up pinnacle as a wayland session:
+    ```nix
+      programs.pinnacle = {
+        enable = true;
+        # ensure portals are installed such that screenshots, screen recording, etc. work properly
+        xdg-portals.enable = true;
+        withUWSM = true;
+      };
+
+      # make sure to set the following if you're using auto-login
+      services.displayManager = {
+        defaultSession = "pinnacle-uwsm";
+      };
+  
+      # not sure if the following is strictly necessary but you probably
+      # want to make sure they're set somewhere:
+      services.xserver.enable = true;
+      programs.dconf.enable = true;
+      security.polkit.enable = true;
+      hardware.graphics.enable = true;
+      # you should also make sure the xkb layout is configured
+    ```
+
+Then we need to configure your users to launch pinnacle properly:
+
+1. load the home-manager module:
+    ```nix
+    {
+      home-manager.sharedModules = [inputs.pinnacle.hmModules.default];
+    }
+    ```
+1. set up the user service/targets in one of your home-manager user modules:
+    ```nix
+      wayland.windowManager.pinnacle = {
+        enable = true;
+        # if you're using rust -- if you're using lua, you'll need to set `wayland.windowManager.pinnacle.config.execCmd` to point to
+        # your lua script.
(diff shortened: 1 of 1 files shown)


nix/modules/home-manager/default.nix
```diff
@@ -0,0 +1,158 @@
...
+    };
+  };
+
+  config = let
+    configFile = settingsFormat.generate "pinnacle.toml" cfg.mergedSettings;
+  in lib.mkIf cfg.enable {
+    xdg.configFile."pinnacle/pinnacle.toml" = {
+      source = configFile;
+      # TODO: make pinnacle reload config when this file changes without needing to restart the whole graphical session
+      # example: https://github.com/nix-community/home-manager/blob/2b73c2fcca690b6eca4f520179e54ae760f25d4e/modules/services/window-managers/i3-sway/sway.nix#L726
```
cassandracomar: one quick/easy option to handle this would be to trap SIGHUP in the pinnacle server and reload config when we receive SIGHUP. then we could use `onChange` to run `kill -HUP $PINNACLE_PID`. we'd need to find the pid but that's potentially easier than adding a msg interface to the pinnacle CLI.

Ottatop: If the lua is set up properly you can do `pinnacle client -e "Pinnacle.reload_config()"`

cassandracomar: cool, got this working. I bumped the versions of lua-protobuf and compat53 to match what's available in nixpkgs. I can undo that and provide source replaced versions of these packages if necessary but it adds complexity.

Ottatop:
Thank you!

PR #358, opened 2025-09-15
feat: implement ext-workspace-v1

cassandracomar:
I've followed the niri approach - predominantly - here, while also using the cosmic implementation as a reference. this PR has been tested with waybar as the ext-workspace-v1 client.

important notes:
1. in order to preserve tag behavior, we don't automatically deactivate active tags when receiving an activate request on a new tag. most compositors treat workspace activation like our `api::tag::switch_to`. in order to treat tags like workspaces, waybar and other clients will need a config option to issue those deactivate on active tags, in concert with the activation request. we can maybe update our own config option as part of the API that changes the behavior of activate to use `api::tag::switch_to` instead of `api::tag::set_active`.
2. I manually set up a test client to check workspace listing and issuing activates/deactivates but didn't get as far as adding test cases. I can complete that work if desired.

### DIFF
--- a/src/handlers/ext_workspace.rs
+++ b/src/handlers/ext_workspace.rs
@@ -1,0 +1,35 @@
+use crate::delegate_ext_workspace;
+use crate::output::OutputName;
+use crate::protocol::ext_workspace::{ExtWorkspaceHandler, ExtWorkspaceManagerState};
+use crate::state::State;
+use crate::tag::TagId;
+
+use smithay::output::Output;
+
+impl ExtWorkspaceHandler for State {
+    fn ext_workspace_manager_state(&mut self) -> &mut ExtWorkspaceManagerState {
+        &mut self.pinnacle.ext_workspace_state
+    }
+
+    fn activate_workspace(&mut self, id: TagId) {
+        if let Some(tag) = id.tag(&self.pinnacle) {
+            crate::api::tag::set_active(self, &tag, Some(true));
+        }
+    }
+
+    fn deactivate_workspace(&mut self, id: TagId) {
+        if let Some(tag) = id.tag(&self.pinnacle) {
+            crate::api::tag::set_active(self, &tag, Some(false));
+        }
+    }
+
+    fn remove_workspace(&mut self, id: TagId) {
+        crate::api::tag::remove(self, Vec::from_iter(id.tag(&self.pinnacle)));
+    }
+
+    fn add_workspace(&mut self, output: &Output, name: String) {
+        crate::api::tag::add(self, [name.clone()], OutputName(output.name()));
+    }
+}
+
+delegate_ext_workspace!(State);
--- a/src/protocol/ext_workspace.rs
+++ b/src/protocol/ext_workspace.rs
@@ -1,0 +1,623 @@
+use std::collections::HashMap;
+use std::collections::hash_map::Entry;
+use std::mem;
+
+use ext_workspace_group_handle_v1::ExtWorkspaceGroupHandleV1;
+use ext_workspace_handle_v1::ExtWorkspaceHandleV1;
+use ext_workspace_manager_v1::ExtWorkspaceManagerV1;
+use smithay::output::Output;
+use smithay::reexports::wayland_protocols::ext::workspace::v1::server::{
+    ext_workspace_group_handle_v1, ext_workspace_handle_v1, ext_workspace_manager_v1,
+};
+use smithay::reexports::wayland_server::protocol::wl_output::WlOutput;
+use smithay::reexports::wayland_server::{
+    Client, DataInit, Dispatch, DisplayHandle, GlobalDispatch, New, Resource,
+};
+use wayland_backend::server::ClientId;
+
+use crate::state::{State, WithState};
+use crate::tag::{Tag, TagId};
+
+const VERSION: u32 = 1;
+
+pub trait ExtWorkspaceHandler {
+    fn ext_workspace_manager_state(&mut self) -> &mut ExtWorkspaceManagerState;
+    fn activate_workspace(&mut self, id: TagId);
+    fn deactivate_workspace(&mut self, id: TagId);
+    fn remove_workspace(&mut self, id: TagId);
+    fn add_workspace(&mut self, output: &Output, name: String);
+}
+
+enum Action {
+    Activate(TagId),
+    Deactivate(TagId),
+    Remove(TagId),
+}
+
+impl Action {
+    fn order(&self) -> u8 {
+        match self {
+            Action::Activate(_) => 2,
+            Action::Deactivate(_) => 1,
+            Action::Remove(_) => 0,
+        }
+    }
+}
+
+pub struct ExtWorkspaceManagerState {
+    display: DisplayHandle,
+    instances: HashMap<ExtWorkspaceManagerV1, Vec<Action>>,
+    tag_groups: HashMap<Output, ExtWorkspaceGroupData>,
+    tags: HashMap<TagId, ExtWorkspaceData>,
+}
+
+struct ExtWorkspaceGroupData {
+    instances: Vec<ExtWorkspaceGroupHandleV1>,
+}
+
+struct ExtWorkspaceData {
+    id: String,
+    name: String,
+    state: ext_workspace_handle_v1::State,
+    instances: Vec<ExtWorkspaceHandleV1>,
+    output: Output,
+}
+
+pub struct ExtWorkspaceGlobalData {
+    filter: Box<dyn for<'c> Fn(&'c Client) -> bool + Send + Sync>,
+}
+
+fn tags_by_output<'a>(
+    outputs: impl Iterator<Item = &'a Output> + 'a,
+) -> impl Iterator<Item = (Output, Tag)> {
+    outputs.flat_map(|output| {
+        output.with_state(|outp| {
+            outp.tags
+                .clone()
+                .into_iter()
+                .map(|tag| (output.clone(), tag))
+        })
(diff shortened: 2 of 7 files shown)


cassandracomar:
I think that was opened after I started working on this?

cassandracomar:
appreciated!

cassandracomar:
per [this comment](https://github.com/pinnacle-comp/pinnacle/issues/359#issuecomment-3313249089) I've removed `Assign` support and added `Remove` support.

xMAC94x:
@cassandracomar the function `on_output_bound` is pub till all the way up the crate, IIRC rust won't give dead_code in that case.
(I didn't saw you on discord, tell me if you are there, because I dont really want to spam on github a lot).

I tried to verify this PR with `wayland` but had no luck so far. I got it to call `GlobalDispatch<ExtWorkspaceManagerV1, ExtWorkspaceGlobalData, D>`  and continue testing soon :)

cassandracomar:
> @cassandracomar the function `on_output_bound` is pub till all the way up the crate, IIRC rust won't give dead_code in that case. (I didn't saw you on discord, tell me if you are there, because I dont really want to spam on github a lot).

sorry, I'm not on discord -- is there a matrix/irc channel I should pop into?

> I tried to verify this PR with `wayland` but had no luck so far. I got it to call `GlobalDispatch<ExtWorkspaceManagerV1, ExtWorkspaceGlobalData, D>` and continue testing soon :)

I'm still checking all the details and need to figure out how to set up a direct test to validate that everything is in place. I /think/ we've done everything necessary here to expose support for the extension but I haven't actually validated that yet. once I've done so, I'll undraft the PR.

Ph4ntomas:
> > @cassandracomar the function `on_output_bound` is pub till all the way up the crate, IIRC rust won't give dead_code in that case. (I didn't saw you on discord, tell me if you are there, because I dont really want to spam on github a lot).
> 
> sorry, I'm not on discord -- is there a matrix/irc channel I should pop into?
> 

There is a matrix space with several channels, and everything is bridged to discord:
https://matrix.to/#/#pinnacle:matrix.org

Ph4ntomas:
> I'm still checking all the details and need to figure out how to set up a direct test to validate that everything is in place.

You might be able to extend the integration test client to support the new protocol, and populate some variables you can the check in one or several test. Ideally we'd want to have a way to setup expectations ahead of time to check for events, but the support is not implemented

For server-side side-effects, you can probe the server internals from the test-suite.

I did some protocol validation in this PR, maybe it can give you some idea as to how to test this one :)

https://github.com/pinnacle-comp/pinnacle/pull/349

xMAC94x:
Just wanna confirm, with `waybar` the latest changes on this PR is working and creating the right visuals. I used the `"ext/workspaces"` module.

Enabling (switching) a tag (on the focused output) via click does not yet currently work.

my config is:
```
"ext/workspaces": {
    "active-only": true,
    "format": "{name}",
    "persistent-workspaces": {
        "3": [], // Always show a workspace with name '3', on all outputs if it does not exists
        "4": ["eDP-1"], // Always show a workspace with name '4', on output 'eDP-1' if it does not exists
        "5": ["eDP-1", "DP-2"] // Always show a workspace with name '5', on outputs 'eDP-1' and 'DP-2' if it does not exists
    }
  },
  ```
  
  `active-only` works.
  `persistent-workspaces` seem not to work

**Comment by cassandracomar** [comment]
I'm not sure what `persistent-workspaces` is supposed to do on the compositor side. isn't that purely a UI thing?

File: src/protocol/ext_workspace.rs [hunk 1]
```diff
@@ -0,0 +1,620 @@
...
+struct ExtWorkspaceGroupData {
+    instances: Vec<ExtWorkspaceGroupHandleV1>,
+}
+
+struct ExtWorkspaceData {
+    id: String,
+    name: String,
+    state: ext_workspace_handle_v1::State,
+    instances: Vec<ExtWorkspaceHandleV1>,
+    output: Option<Output>,
```
Ottatop: Is there a reason this is an `Option`? If not we should just make it `Output`.

src/protocol/ext_workspace.rs
```diff
@@ -0,0 +1,620 @@
...
+    tag_id: TagId,
+    tag: &Tag,
+) -> bool {
+    let mut state = ext_workspace_handle_v1::State::empty();
+    if tag.active() {
+        state |= ext_workspace_handle_v1::State::Active;
+    }
+    if output.with_state(|outp| outp.focused_tags().any(|ft| ft.id() == tag_id)) {
+        state |= ext_workspace_handle_v1::State::Urgent;
+    }
```
Ottatop: Not sure what the purpose of `tag_id` is here. Currently `refresh_workspace` is always getting called with `.., tag.id(), &tag)` so this always makes active workspace handles urgent, don't think that's intended.

src/protocol/ext_workspace.rs
```diff
@@ -0,0 +1,620 @@
...
+        Entry::Occupied(entry) => {
+            // Existing workspace, check if anything changed.
+            let data = entry.into_mut();
+
+            let mut state_changed = false;
+            if data.state != state {
+                data.state = state;
+                state_changed = true;
+            }
+            // TODO: do we need to handle tag renames/destroy events?
```
Ottatop: Destroys are handled in `refresh` already, right? As for renames we don't expose an api for that at the moment so we can leave that for the future, unless you'd like to implement it up front here.

src/protocol/ext_workspace.rs
```diff
@@ -0,0 +1,620 @@
...
+        _state: &mut D,
+        _client: &Client,
+        _resource: &ExtWorkspaceGroupHandleV1,
+        request: <ExtWorkspaceGroupHandleV1 as Resource>::Request,
+        _data: &ExtWorkspaceManagerV1,
+        _dhandle: &DisplayHandle,
+        _data_init: &mut DataInit<'_, D>,
+    ) {
+        match request {
+            ext_workspace_group_handle_v1::Request::CreateWorkspace { .. } => (),
```
Ottatop: We should also support creating workspaces here seeing as we can easily add tags at runtime.

Ottatop (commented):
Seems to work well enough, great work! A few notes.

src/protocol/ext_workspace.rs
cassandracomar: it's called on each tag for each output [here](https://github.com/pinnacle-comp/pinnacle/pull/358/files#diff-9917e42a393d43f7131e82aa29beb61c3e7ed50b8a848328b11380fd954cde8cR160). but I agree it doesn't need to be a separate argument to the function as we can get this info via `Tag::id`.

are you actually seeing every tag getting marked as `State::Urgent`? that's a more serious bug that needs investigation if so.

src/protocol/ext_workspace.rs
cassandracomar: yeah, think I just forgot to remove this comment. will clean it up.

src/protocol/ext_workspace.rs
Ph4ntomas: `OutputState::focused_tag()` returns all active tags for the output (https://github.com/pinnacle-comp/pinnacle/blob/4e7aad0d07fd958ff43b6c9b2195391f7f48ead3/src/output.rs#L133-L137).

I not aware of a `urgent client` or `urgent workspace` at the moment.

Ph4ntomas: Well, the only thing I found is a TODO:
https://github.com/pinnacle-comp/pinnacle/blob/4e7aad0d07fd958ff43b6c9b2195391f7f48ead3/src/handlers/xdg_activation.rs#L107

cassandracomar: ahh, it should likely be something akin to "tags with currently focused windows", probably?

Ph4ntomas: I think the 'urgent' stuff should be kept for later.

On X11, window could raise an `urgent` flag to ask for user attention (used e.g. for notifications). AwesomeWM use that mark the tags the window has as urgent so you can see something happened, even if the tag is currently inactive

In my opinion this state should be kept to implement something similar.

cassandracomar: sounds good, removed workspace urgency for now.

src/handlers/ext_workspace.rs
```diff
@@ -0,0 +1,44 @@
...
+            .outputs
+            .iter()
+            .map(Output::name)
+            .map(OutputName)
+            .collect::<Vec<_>>();
+        for output in outputs {
+            crate::api::tag::add(self, [name.clone()], output);
+        }
+    }
+}
```
Ottatop: `ext_workspace_group_handle_v1::create_workspace` specifies that the workspace should be created on the group, so this needs to take the output for the group.

src/protocol/ext_workspace.rs
```diff
@@ -0,0 +1,614 @@
...
+        let group = client
+            .create_resource::<ExtWorkspaceGroupHandleV1, _, D>(
+                handle,
+                manager.version(),
+                manager.clone(),
+            )
+            .unwrap();
+        manager.workspace_group(&group);
+
+        group.capabilities(ext_workspace_group_handle_v1::GroupCapabilities::empty());
```
Ottatop: Group capabilities should have `CreateWorkspace`.

src/handlers/ext_workspace.rs
cassandracomar: ahh I missed that, sorry. fixed it to only create the workspace on the matching output for the group.

PR #360, opened 2025-09-17
Adds send_message method to LayerHandle

Ph4ntomas:
The goal of this method is to allow the refresh of a layer & its widgets


Ottatop (approved):
Lgtm, thank you!

PR #362, opened 2025-09-20
Add output enable/disable signal and change the current ones

anriha:
This PR does two things. 
### Part 1: Changes to Output signals
It changes output signals for the api. Currently there are OutputConnect and OutputDisconnect. But these didn't really do what they should. Connect was only fired when output was connected for the first time, so there was no way to get signal when output was activated. And disconnect fired every time the monitor was switched off in any way, not just when it was actually disconnected. This fixes this behavior and adds two new signals OutputEnable and OutputDisable. 

### Part 2: Fix Silent Output Reactivation on Hotplug

This commit resolves an issue where an output could be re-enabled by the backend without firing the appropriate signals.

**Motivation:**
I was observing an issue where one of my monitors would power back on a few seconds after being disabled by an idle manager. This behavior was also present in other compositors like river and dwl, suggesting a backend or driver-level quirk that the compositor should be resilient to. Strangely enough it didn't happen in sway.

My initial attempt to fix this in my config by listening for an "enable" signal failed, as no signal was being fired when the monitor reawakened.

**Root Cause:**
Investigation revealed that this reactivation scenario bypasses the standard `set_output_enabled` and `set_output_powered` code paths. The backend handles the underlying hotplug event by calling `change_output_state` directly to apply a mode. This action physically powered on the monitor but did not update the compositor's logical state or fire the `OutputEnable` signal, leading to a desynchronization.

**Solution:**
To resolve this, the following precondition has been added to the top of `change_output_state`:

```rust
if mode.is_some() {
    self.set_output_enabled(output, true);
}
```
This change makes sure that setting a mode on an output is an explicit act of enabling it. By ensuring set_output_enabled(true) is called first, we guarantee that all code paths that physically activate a monitor are funneled through the correct signaling logic. This keeps the compositor's state consistent with the hardware and resolves the bug.

### DIFF
--- a/src/output.rs
+++ b/src/output.rs
@@ -308,23 +308,17 @@ impl Pinnacle {
             self.space.map_output(output, output.current_location());
 
             // Trigger the connect signal here for configs to reposition outputs
-            //
-            // TODO: Create a new output_disable/enable signal and trigger it here
-            // instead of connect and disconnect
             if should_signal {
                 self.signal_state.output_connect.signal(output);
             }
         } else {
             if let Some(global) = output.with_state_mut(|state| state.enabled_global_id.take()) {
                 self.display_handle.remove_global::<State>(global);
+
+                // Trigger the disconnect signal here for configs to reposition outputs
+                self.signal_state.output_disconnect.signal(output);
             }
             self.space.unmap_output(output);
-
-            // Trigger the disconnect signal here for configs to reposition outputs
-            //
-            // TODO: Create a new output_disable/enable signal and trigger it here
-            // instead of connect and disconnect
-            self.signal_state.output_disconnect.signal(output);
 
             self.gamma_control_manager_state.output_removed(output);
 
--- a/api/lua/pinnacle/output.lua
+++ b/api/lua/pinnacle/output.lua
@@ -140,6 +140,7 @@ end
 local signal_name_to_SignalName = {
     connect = "OutputConnect",
     disconnect = "OutputDisconnect",
+    setup = "OutputSetup",
     resize = "OutputResize",
     move = "OutputMove",
     pointer_enter = "OutputPointerEnter",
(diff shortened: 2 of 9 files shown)


src/backend.rs
```diff
@@ -187,6 +188,12 @@ impl State {
         self.pinnacle
             .output_power_management_state
             .mode_set(output, powered);
+
+        if powered {
+            self.pinnacle.signal_state.output_enable.signal(output);
+        } else {
+            self.pinnacle.signal_state.output_disable.signal(output);
+        }
```
Ottatop: Powered != enable/disable, so this should be removed.

src/output.rs
```diff
@@ -308,23 +312,19 @@ impl Pinnacle {
...
 
             // Trigger the connect signal here for configs to reposition outputs
-            //
-            // TODO: Create a new output_disable/enable signal and trigger it here
-            // instead of connect and disconnect
             if should_signal {
                 self.signal_state.output_connect.signal(output);
             }
+
+            self.signal_state.output_enable.signal(output);
```
Ottatop: This should be in the `if should_signal` block above so it doesn't get triggered again if you enable an already enabled output.

src/output.rs
```diff
@@ -308,23 +312,19 @@ impl Pinnacle {
...
             }
             self.space.unmap_output(output);
 
-            // Trigger the disconnect signal here for configs to reposition outputs
-            //
-            // TODO: Create a new output_disable/enable signal and trigger it here
-            // instead of connect and disconnect
-            self.signal_state.output_disconnect.signal(output);
+            // Trigger the disable signal here for configs to reposition outputs
+            self.signal_state.output_disable.signal(output);
```
Ottatop: This should be in the if block above for the same reason above.

src/output.rs
```diff
@@ -179,6 +179,10 @@ impl Pinnacle {
     ) {
         let _span = tracy_client::span!("Pinnacle::change_output_state");
 
+        if let Some(_mode) = mode {
```
Ottatop:
```suggestion
        if mode.is_some() {
```

src/output.rs
```diff
@@ -179,6 +179,10 @@ impl Pinnacle {
...
+            self.set_output_enabled(output, true);
```
Ottatop: I'm not a fan of this being here. If the output was explicitly disabled I don't think I'd want a mode change to re-enable it; I'd expect it to stay disabled.

anriha:
> Instead of enable/disable signals, I think we should instead fix OutputSignal::connect so that it fires correctly on all connections (though since you've already done some work here I guess I could merge these signals after some fixes). However this would require some changes to output state persistence.

This does make sense. I was mainly assuming this enable/disable signal is what you intended because of this comment;
```rust
// TODO: Create a new output_disable/enable signal and trigger it 
// instead of connect and disconnect
```

But, having connect fire every time, we need it to would make sense, too, especially if there was a setup command or something like that.

anriha:
But you are right that 
```rust
        if let Some(_mode) = mode {
            self.set_output_enabled(output, true);
```
Shouldn't be there. I will debug things more to really find out what is going here.

PR #364, opened 2025-09-21
snowcap: add mouse_area widget

Ph4ntomas:
This PR adds the MouseArea widget to snowcap.

This widget is a native Iced widget that allows finer control compared to Button, (e.g. knowing which mouse button has been pressed).

TODO:
- [] Correctly handle scroll event
- [] Correctly handle move event


Ph4ntomas:
Side note about the unique_id: It might be useful even for widget without callback.

One of the reason I need the mouse area is to properly handle a clickable taglist for my bar. I'll likely use that value as an identifier for the state in the bar itself (I wrapped widgets so they have a view & update function), and that may prove useful for other stateful widgets, even without events driven by snowcap (e.g. a textbox)

Ph4ntomas:
So the trick I was using in lua doesn't feel rust-y enough for my taste (memoize some value in a map/dictionary).

I'm adding the unique_id to the GetWidgetEventsResponse type for the time being, but I really think it would be best to have it inside a `message Key {}` type (I just don't want to do API breaking changes without permission :) )

snowcap/api/protobuf/snowcap/widget/v1/widget.proto
```diff
@@ -290,6 +291,90 @@ message Image {
...
+  bool on_right_release = 6;
+  bool on_middle_press = 7;
+  bool on_middle_release = 8;
+  bool on_scroll = 9;
+  bool on_enter = 10;
+  bool on_move = 11;
+  bool on_exit = 12;
+  optional Interaction interaction = 13;
+  optional string unique_id = 14;
+  optional uint32 widget_id = 15;
```
Ph4ntomas: I feel like these two could be merged. There are other iced widget that might need a string as an id (e.g. `TextArea` and `TextEditor`). 

I went into more depth in https://github.com/pinnacle-comp/pinnacle/pull/364#issuecomment-3330665771 as to why this is needed at least in `MouseArea`, but the gist of it is that without a way to identify this widget accross `view()` generations, the double click event message tied to the `widget_id` is dropped because a new view is created between the second `press` and the `double_click` handling in the config (the two event are sent at the same time).

snowcap/api/rust/src/widget.rs
```diff
@@ -51,6 +53,35 @@ impl From<u32> for WidgetId {
...
+pub(crate) enum WidgetKey {
+    WidgetId(WidgetId),
+    UniqueId(String),
+}
+
+impl Default for WidgetKey {
+    fn default() -> Self {
+        Self::WidgetId(Default::default())
+    }
+}
```
Ph4ntomas: re merging `unique_id` and `widget_id` as a new protobuf `message`:
At least on the rust side, this would mean the internal repr and the protobuf one would be coherent.

snowcap/api/lua/snowcap/layer.lua
```diff
@@ -130,28 +138,37 @@ function layer.new_widget(args)
...
-                    widget_def,
-                    callbacks,
-                    function(callbacks, widget)
-                        if widget.button and widget.button.on_press then
-                            callbacks[widget.button.widget_id] = widget.button.on_press
-                        end
-                    end
+            msg = callbacks[widget_id]
+        elseif response.mouse_area then
+            widget_id = response.mouse_area.unique_id or widget_id
```
Ph4ntomas: Here, having the `unique_id` + `widget_id` unified `message` would mage it less weird, since you wouldn't need to first check what kind of message the response contain it is before using it.

As a side note:
It's possible to define `callbacks` to accept a table as key and do the right thing by setting `__index` and `__newindex` as functions in its metatable.

 However, I think it's ok to just extract the 'correct one' depending on whether the `unique_id` is set or not, and I don't really like to do that kind of `metatable` magic unless absolutely needed.

Ottatop:
On the second click, `MouseArea` will generate both click and double click events within a single update. I think it's better to have `GetWidgetEventsResponse` send a vec of events so both can be sent at the same time, which would allow the client to apply `view` twice without having to call `update` in between.

I'm not a fan of exposing a user-facing ID method and I think the current system of auto-generating IDs is good because we can hide backend implementation details like how widget IDs work. Exposing a unique_id method and requiring people to provide a unique ID is not ideal because we are a) pushing implementation details into the user-facing API which limits the ability to make breaking changes, and b) is error-prone. The current MouseArea API needs `unique_id` to be called when both `on_press` and `on_double_click`, but nothing actually enforces that. Someone could just forget to add it and wonder why the double click doesn't work, which is a sharp corner in the API that I don't want.

I'd do away with the unique ID and implement my suggestion above instead. This means we won't need the user to call a separate method to make things work.

> I just don't want to do API breaking changes without permission

I believe widget events were added after the most recent release so you don't have to worry about breaking changes in the development window since then.

Ph4ntomas:
> On the second click, MouseArea will generate both click and double click events within a single update. I think it's better to have GetWidgetEventsResponse send a vec of events so both can be sent at the same time, which would allow the client to apply view twice without having to call update in between.

> I think the current system of auto-generating IDs is good because we can hide backend implementation details like how widget IDs work.

While I agree it's best to hide implementation details, this is causing some events to be dropped, even when batching events in GetWidgetEventsResponse (although that does fix the double click). The issue is that surface::update call can be called more than once for a given view generation, but the client will always regenerate a view if at least on event (or event batch) is received.

Added some log in snowcap_api::layer.rs to test this:

Move event are sometime dropped: 
```
2025-09-28T09:11:33.598709Z  INFO pinnacle::config: UPDATE Move(
2025-09-28T09:11:33.598757Z  INFO pinnacle::config:     Point {
2025-09-28T09:11:33.598783Z  INFO pinnacle::config:         x: 176.21484,
2025-09-28T09:11:33.598806Z  INFO pinnacle::config:         y: 87.203125,
2025-09-28T09:11:33.598830Z  INFO pinnacle::config:     },
2025-09-28T09:11:33.598852Z  INFO pinnacle::config: )
2025-09-28T09:11:33.598880Z  INFO pinnacle::config: view called
2025-09-28T09:11:33.694117Z  INFO pinnacle::config: UPDATE Move(
2025-09-28T09:11:33.694331Z  INFO pinnacle::config:     Point {
2025-09-28T09:11:33.694464Z  INFO pinnacle::config:         x: 176.89063,
2025-09-28T09:11:33.694585Z  INFO pinnacle::config:         y: 87.203125,
2025-09-28T09:11:33.694695Z  INFO pinnacle::config:     },
2025-09-28T09:11:33.694821Z  INFO pinnacle::config: )
2025-09-28T09:11:33.694941Z  INFO pinnacle::config: view called
2025-09-28T09:06:40.326876Z  INFO pinnacle::config: NOMSG WidgetId(
2025-09-28T09:06:40.326943Z  INFO pinnacle::config:     113,
2025-09-28T09:06:40.326997Z  INFO pinnacle::config: ) => MouseArea(
2025-09-28T09:06:40.327048Z  INFO pinnacle::config:     Event {
2025-09-28T09:06:40.327108Z  INFO pinnacle::config:         event_type: EventExit,
2025-09-28T09:06:40.327174Z  INFO pinnacle::config:         data: None,
2025-09-28T09:06:40.327231Z  INFO pinnacle::config:     },
2025-09-28T09:06:40.327320Z  INFO pinnacle::config: )
2025-09-28T09:06:40.327415Z  INFO pinnacle::config: avail key => [WidgetId(114)]
```

Scroll events are particularly bad:
```
2025-09-28T09:49:23.228364Z  INFO pinnacle::config: UPDATE Scroll(
2025-09-28T09:49:23.228395Z  INFO pinnacle::config:     Lines {
2025-09-28T09:49:23.228408Z  INFO pinnacle::config:         x: 0.0,
2025-09-28T09:49:23.228421Z  INFO pinnacle::config:         y: -1.0,
2025-09-28T09:49:23.228435Z  INFO pinnacle::config:     },
2025-09-28T09:49:23.228447Z  INFO pinnacle::config: )
2025-09-28T09:49:23.228461Z  INFO pinnacle::config: view called
2025-09-28T09:49:23.230134Z  INFO pinnacle::config: dropped event WidgetId(98)
2025-09-28T09:49:23.230166Z  INFO pinnacle::config: avail id: "[WidgetId(99)]"
2025-09-28T09:49:23.230181Z  INFO pinnacle::config: dropped event WidgetId(98)
2025-09-28T09:49:23.230198Z  INFO pinnacle::config: avail id: "[WidgetId(99)]"
2025-09-28T09:49:23.230214Z  INFO pinnacle::config: dropped event WidgetId(98)
2025-09-28T09:49:23.230232Z  INFO pinnacle::config: avail id: "[WidgetId(99)]"
2025-09-28T09:49:23.273369Z  INFO pinnacle::config: UPDATE Scroll(
2025-09-28T09:49:23.273435Z  INFO pinnacle::config:     Lines {
2025-09-28T09:49:23.273468Z  INFO pinnacle::config:         x: 0.0,
2025-09-28T09:49:23.273498Z  INFO pinnacle::config:         y: -1.0,
2025-09-28T09:49:23.273530Z  INFO pinnacle::config:     },
2025-09-28T09:49:23.273560Z  INFO pinnacle::config: )
2025-09-28T09:49:23.273596Z  INFO pinnacle::config: view called
2025-09-28T09:49:23.277340Z  INFO pinnacle::config: dropped event WidgetId(99)
2025-09-28T09:49:23.277395Z  INFO pinnacle::config: avail id: "[WidgetId(100)]"
2025-09-28T09:49:23.277422Z  INFO pinnacle::config: dropped event WidgetId(99)
2025-09-28T09:49:23.277447Z  INFO pinnacle::config: avail id: "[WidgetId(100)]"
2025-09-28T09:49:23.277472Z  INFO pinnacle::config: dropped event WidgetId(99)
2025-09-28T09:49:23.277498Z  INFO pinnacle::config: avail id: "[WidgetId(100)]"
2025-09-28T09:49:23.277525Z  INFO pinnacle::config: dropped event WidgetId(99)
2025-09-28T09:49:23.277552Z  INFO pinnacle::config: avail id: "[WidgetId(100)]"
2025-09-28T09:49:23.277582Z  INFO pinnacle::config: dropped event WidgetId(99)
2025-09-28T09:49:23.277622Z  INFO pinnacle::config: avail id: "[WidgetId(100)]"
2025-09-28T09:49:23.277652Z  INFO pinnacle::config: dropped event WidgetId(99)
2025-09-28T09:49:23.277684Z  INFO pinnacle::config: avail id: "[WidgetId(100)]"
2025-09-28T09:49:23.277715Z  INFO pinnacle::config: dropped event WidgetId(99)
2025-09-28T09:49:23.277746Z  INFO pinnacle::config: avail id: "[WidgetId(100)]"
2025-09-28T09:49:23.277775Z  INFO pinnacle::config: dropped event WidgetId(99)
2025-09-28T09:49:23.277806Z  INFO pinnacle::config: avail id: "[WidgetId(100)]"
2025-09-28T09:49:23.277835Z  INFO pinnacle::config: dropped event WidgetId(99)
2025-09-28T09:49:23.277869Z  INFO pinnacle::config: avail id: "[WidgetId(100)]"
2025-09-28T09:49:23.288885Z  INFO pinnacle::config: UPDATE Scroll(
2025-09-28T09:49:23.288985Z  INFO pinnacle::config:     Lines {
2025-09-28T09:49:23.289076Z  INFO pinnacle::config:         x: 0.0,
2025-09-28T09:49:23.289197Z  INFO pinnacle::config:         y: -1.0,
2025-09-28T09:49:23.289283Z  INFO pinnacle::config:     },
2025-09-28T09:49:23.289356Z  INFO pinnacle::config: )
2025-09-28T09:49:23.289435Z  INFO pinnacle::config: view called
2025-09-28T09:49:23.291875Z  INFO pinnacle::config: dropped event WidgetId(100)
2025-09-28T09:49:23.291968Z  INFO pinnacle::config: avail id: "[WidgetId(101)]"
2025-09-28T09:49:23.292041Z  INFO pinnacle::config: dropped event WidgetId(100)
2025-09-28T09:49:23.292108Z  INFO pinnacle::config: avail id: "[WidgetId(101)]"
2025-09-28T09:49:23.292195Z  INFO pinnacle::config: dropped event WidgetId(100)
2025-09-28T09:49:23.292266Z  INFO pinnacle::config: avail id: "[WidgetId(101)]"
2025-09-28T09:49:23.299020Z  INFO pinnacle::config: UPDATE Scroll(
2025-09-28T09:49:23.299061Z  INFO pinnacle::config:     Lines {
2025-09-28T09:49:23.299084Z  INFO pinnacle::config:         x: 0.0,
2025-09-28T09:49:23.299101Z  INFO pinnacle::config:         y: -1.0,
2025-09-28T09:49:23.299125Z  INFO pinnacle::config:     },
2025-09-28T09:49:23.299143Z  INFO pinnacle::config: )
2025-09-28T09:49:23.299179Z  INFO pinnacle::config: UPDATE Scroll(
2025-09-28T09:49:23.299201Z  INFO pinnacle::config:     Lines {
2025-09-28T09:49:23.299224Z  INFO pinnacle::config:         x: 0.0,
2025-09-28T09:49:23.299247Z  INFO pinnacle::config:         y: -1.0,
2025-09-28T09:49:23.299270Z  INFO pinnacle::config:     },
2025-09-28T09:49:23.299299Z  INFO pinnacle::config: )
2025-09-28T09:49:23.299332Z  INFO pinnacle::config: view called
```

Here is the specific code I use for testing:
[testing.patch](https://github.com/user-attachments/files/22581855/testing.patch)

apply the patch, then run:
`just run -c ./api/rust/examples/default_config`

Ph4ntomas:
This is ultimately your call, but the way I see it:
1. We decide that some event drop are acceptable
2. We expose a way for user to have either a unique key across view or an autogenerated id, and document that some widget may drop events if the autogenerated id is used (i.e. it's there for convenience/ease of use, and if reliability is needed, best use the unique key)
3. We associate the widget_id per layer, and reset it before each view generation => This should ensure that unless the order/amount of widget changes, the widget_ids remain stable accross generation.
4. We generate some kind of hash as a widget_id (this require Msg to implement Hash and I'm not sure this can be done with closure, so ... meh)

I'm not a huge fan of `#1`.
`#3` could cause issue if the order & amount of widget changes, but the widget type stays the same. That being said this is a very niche issue, and might never appear in the wild :)

Regardless, I think the event batching is a good idea :D

I'm of course open to suggestion if you have a better idea.

Ph4ntomas:
Uh I've found a slight issue that only affect winit. 

Mouse motion event's can get *very* chatty and can overwhelm iced_futures subscriptions channels (which have a size of 100 events AFAICT). If this happens, any further events in the batch are dropped which is less than ideal (if anything, I'd rather drop mouse events than other potentially more useful ones)

A secondary issue is that the spammy nature of mouse motion events can lead to a bit of lag (that being said, I've compiled in debug, and I've added logging. Removing the logs themselves lead to fewer warning from iced, but the config taking too much time to handle events and send a view could cause events to accumulate).

This is tied to the monitor/window framerate (main monitor refreshes @240 fps and have no issue whatsoever). I barely hit ~50 evts per update). Secondary monitor @60fps and can reach ~150-200 if I wiggle the mouse.

I see a few possible mitigation:
- throttle the mouse moved events. The raw events contain the 'time' (in milisec as far as I can tell) so we could throttle that way
- discard mouse_move events before we broadcast events to subscriptions but after the widget handles them -> This means we can't have subscriptions on mouse move (we don't have any at the moment), but the widgets themselves gets all events. This doesn't mitigate the potential lag issue.

In my opinion, if this is to be addressed, throttling motion events should be the way to go. 
That being said, the issue only occurs on winit backend (max I've seen on udev is 14 events, and this is consistent regardless of the monitor), so I'm not even sure it's worth mitigating it.

Ph4ntomas:
> That being said, the issue only occurs on winit backend (max I've seen on udev is 14 events, and this is consistent regardless of the monitor), so I'm not even sure it's worth mitigating it.

I just want to clarify this. After further testing, I'm still unable to reproduce when pinnacle is started from tty (regardless of compilation option).
With the winit backend, I have way less lag/warnings if I remove all logging.

I *do* think this can & should be mitigated by throttling mouse-move events (or any kind of chatty/repetitive hw events).
However, since this is tied to winit, I'd rather open an issue to take some time to think about a mitigation scheme that wouldn't impact udev backend. Instead of 'not worth mitigating it', I meant 'not worth mitigating it right now, in this PR'

Ideally it would be some dynamic adjustment to prevent hw events from overwhelming iced channels, while still allowing most events to go through. 
As an example of such scheme: if the event_queue is regularly filled above 75% or has reached the threshold at least once, we increase the throttling delay.
This could also be achieved by increasing channel size a smidge in iced, but afaik they don't expose any way to do that at the moment

snowcap/api/protobuf/snowcap/widget/v1/widget.proto
```diff
@@ -290,21 +291,108 @@ message Image {
...
+    EVENT_DOUBLE_CLICK = 2;
+    EVENT_RIGHT_PRESS = 3;
+    EVENT_RIGHT_RELEASE = 4;
+    EVENT_MIDDLE_PRESS = 5;
+    EVENT_MIDDLE_RELEASE = 6;
+    EVENT_SCROLL = 7;
+    EVENT_ENTER = 8;
+    EVENT_MOVE = 9;
+    EVENT_EXIT = 10;
+  }
```
Ottatop: Enum names should follow {ENUM_NAME_IN_SNAKECASE}\_{VARIANT_NAME}, so these should be EVENT_TYPE_*

snowcap/api/protobuf/snowcap/widget/v1/widget.proto
```diff
@@ -290,21 +291,108 @@ message Image {
...
+
+  message Event {
+    EventType event_type = 1;
+    oneof data {
+      ScrollDelta scroll_delta = 2;
+      Point point = 3;
+    }
+  }
```
Ottatop, line 372: I'd prefer this to be just a oneof with all the unit variants being `google.protobuf.Empty` instead of separating the data from the enum.

snowcap/api/lua/snowcap/widget.lua
```diff
@@ -172,6 +173,86 @@ local content_fit = {
 ---@field height snowcap.widget.Length?
 ---@field child snowcap.widget.WidgetDef
 
+---Emits messages on mouse events.
+---@class snowcap.widget.MouseArea
+---@field child snowcap.widget.WidgetDef MouseArea content
+---@field callbacks snowcap.widget.mouse_area.Callbacks Set of messages to emit on a mouse event.
```
Ottatop: I'd like to flatten this field out.

Ph4ntomas:
> I'll look at this more in depth soonTM but the current objective is to bugfix and stabilize 0.2.

Does that mean this PR won't make the cut for 0.2 ?

Just to clarify, I don't necessarilly mind the MouseArea not making the cut, but this PR fundamentally change the way we handle events (they are batched, and the surface buffers events until after we receive the next view).

Since this is quite the breaking change I was wondering if it was possible to cherry-pick these two commit for 0.2 and have a 'clean slate' to add widgets in 0.3.

Let me know if you want me to open a PR with just the event handling change.

snowcap/api/lua/snowcap/widget.lua
Ph4ntomas: I don't mind flattening the callbacks field in the MouseArea because it's user facing, but I'd like to keep the Callbacks type for the processing of event. Would that be acceptable ?

snowcap/api/protobuf/snowcap/widget/v1/widget.proto
Ph4ntomas: Make sense. Then I'm removing the `EventType` altogether. 
Also renamed `ScrollDelta` to `ScrollEvent`, and `Point` to `MoveEvent`

snowcap/api/protobuf/snowcap/widget/v1/widget.proto
Ph4ntomas: Closing this, since I've removed the whole Enum :)

snowcap/api/lua/snowcap/widget.lua
Ottatop: Yep! I just don't want people to have to nest callbacks in another table

PR #366, opened 2025-09-28
Add powered to ConnectorSavedState

anriha:
This PR introduces the powered field to `ConnectorSavedState` to correctly handle monitors that automatically try to reconnect after being powered off.

When an output is deactivated, its last known powered state is now saved. Upon reconnection, this saved state is reapplied, ensuring that an output intended to be off remains off. This resolves power-looping issues as discussed in [#362](https://github.com/pinnacle-comp/pinnacle/pull/362).

~Additionally, this change removes the creation of `ConnectorSavedState` from the `set_loc` API call. As far as I can tell there is no situation where that would be used, as it would always be overwritten by either `set_enable` or `remove_output`.~

### DIFF
--- a/src/backend/udev.rs
+++ b/src/backend/udev.rs
@@ -1065,9 +1065,18 @@ impl Udev {
             .connector_saved_states
             .get(&OutputName(output.name()))
         {
-            let ConnectorSavedState { loc, tags, scale } = saved_state;
-            output.with_state_mut(|state| state.tags.clone_from(tags));
-            pinnacle.change_output_state(self, &output, None, None, *scale, Some(*loc));
+            let ConnectorSavedState {
+                loc,
+                tags,
+                scale,
+                powered,
+            } = saved_state.clone();
+
+            output.with_state_mut(|state| state.tags.clone_from(&tags));
+            pinnacle.change_output_state(self, &output, None, None, scale, Some(loc));
+            if let Some(powered) = powered {
+                self.set_output_powered(&output, &pinnacle.loop_handle, powered);
+            }
         } else {
             pinnacle.signal_state.output_connect.signal(&output);
         }
--- a/src/config.rs
+++ b/src/config.rs
@@ -296,6 +296,8 @@ pub struct ConnectorSavedState {
     pub tags: IndexSet<Tag>,
     /// The output's previous scale
     pub scale: Option<smithay::output::Scale>,
+    /// The output's previous powered state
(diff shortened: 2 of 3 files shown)


src/api/output/v1.rs
```diff
@@ -57,24 +56,6 @@ impl output::v1::output_service_server::OutputService for super::OutputService {
...
-            } else {
-                state.pinnacle.config.connector_saved_states.insert(
-                    output_name.clone(),
-                    ConnectorSavedState {
-                        loc: (x, y).into(),
-                        ..Default::default()
-                    },
-                );
-            }
-
```
Ph4ntomas, line 77: Are you sure this is dead code ?

As I understand it, the iftrue branch allows you to set the location of a previously connected output, and the else branch allows you to set the location of an output before it's plugged. 
This can be useful on e.g. a laptop if you already know which port correspond to which name.

anriha: Sorry, what I meant is, that it is not dead code as in it never gets called. It is dead code in a way that it doesn't change anything. As far as I can tell there are only two ways an output can be disabled and that is via the `set_enabled(false)` and `remove_output`. Both of these completely overwrite what is stored in `connector_saved_states`. So no matter what was there it will be overwritten when you actually remove the output.

Ph4ntomas: `set_output_enabled` is called from the OutputManagementProtocol handlers, and `remove_output` is called from udev backend (src/backend/udev.rs:1106).

At least `remove_output` checks that the output object is defined before calling `remove_output`, meaning the saved state is not overriden.

Beside that, this state is also used when an output is connected (or connected again). Meaning, even if `remove_output` or `set_output_enabled` did overwrite the saved state, `set_loc` on a disconnected output would still allow to set the output location ahead of a future re-connection, so I'm still not convinced this is dead code (in the sense of its side effect will always be overwritten and can never be useful)

anriha: There is one case that I did miss. With this, you can use set_loc for output that has never been seen before or is currently disconnected. As there is no check that the output in set_loc is actually valid. You still probably couldn't actually use it like that as it is only called on `OutputHandle` via API, so you would need to have valid output, but I guess in that case it still makes sense to keep it.

anriha: I put it back and changed powered to `Option<bool>` so now it will be None if you use `set_loc` for non-existent output.

PR #367, opened 2025-10-02
feat(nix): add nixGL support for non-NixOS

cassandracomar:
home-manager supports setting up nixGL so make use of that in the
home-manager module to enable correctly launching pinnacle wrapped with
the nixGL nixVulkan* wrappers. home-manager also provides the ability to
set up xdg desktop portals, similar to the nixos options, so expose
those options as well for non-NixOS.


Ottatop (approved):
Thanks!

PR #368, opened 2025-10-03
Add window layout state change signals

anriha:
This PR adds signals to track when windows enter or exit different layout modes (tiled, floating, maximized, fullscreen, and spilled). 
When a window changes from one layout mode to another, the appropriate "unset" signal is fired for the previous state, followed by the "set" signal for the new state. This allows configurations to react to both entering and exiting specific window states. The signal is also fired when the window first spawns.


anriha:
You are right, that would be simpler. But I still think it makes sense to add the information about what layout mode it went from. One of my use cases that I plan for this is to add idle inhibitors to windows. Currently, when you are playing wine game via xwayland there is no way to get idle inhibitor on the window. This is usually fine as you are using mouse/keyboard. But if you play a game with just controller, that doesn't register as any activity. So I want to be able to add when window goes fullscreen add idle inhibitor, when it leaves remove it. 

So I think it might make sense to change it to one signal, something like `WindowLayoutModeChanged` which would get the `old_mode: Option<LayoutModeKind>, new_mode: Option<LayoutModeKind>`. If the window was new the `old_mode` would be `None`. And if the window was closing (which I didn't handle at all thinking about it) the `new_mode` would be `None`.

src/window/layout.rs
```diff
@@ -217,6 +218,42 @@ impl Pinnacle {
...
+                    self.signal_state.window_set_maximized.signal(window);
+                }
+                crate::window::window_state::LayoutModeKind::Fullscreen => {
+                    self.signal_state.window_set_fullscreen.signal(window);
+                }
+                crate::window::window_state::LayoutModeKind::Spilled => {
+                    self.signal_state.window_set_spilled.signal(window);
+                }
+            }
+
```
Ph4ntomas: This is not the only place the layout can change.

For example (there might be other instances):
- src/layout.rs:108
- src/api/window.rs:{443, 453, 459, 469}

anriha:
> You don't need to know the old mode for this to work (that doesn't mean it won't be useful for other stuff).

You do need to know when the window stops being fullscreen, so you can remove the idle inhibitor. You could track it internally in your configuration, but that would be more complicated.

src/window/layout.rs
Ph4ntomas: IMO, it's better to handle this from `State::on_event_loop_cycle_completion()` rather than sprinkle the call to `self.signal_state.window_*`.

If you need the old value, it could be held in the window `LayoutMode` until the end of the loop, then from the aforementioned function, you'd iterate over windows to clear this value and send the old/new pair.
If you don't need the old value, just a boolean would work.

This should prevent missing layout changes.

(You may want to wait for Ottatop input on that first)

Ottatop: Yea I think having an `old_layout_mode: LayoutMode`, checking to see if the current mode != the old mode to send the signal, and copying the current mode to the old mode in `on_event_loop_cycle_completion` is the simplest solution.

Ottatop:
I think it's best to just send the new mode for now. If we send both modes then the old has to be an Option and is only None for the first signal, which I don't like.

anriha:
Ok, in that case, how about this approach. I do need to know what is the window initial layout mode when it spawns, I need to know when it changes, and when it closes. So how about I add new signals, something like window created/destroyed. These would just give you `window_id` and you can then use API to get its layout mode. And I do just add  `WindowLayoutModeChanged`. 

For my use case, this would make it a bit more complicated to keep track of everything in my config, but it would be doable and it would address the issue.  

Theoretically we wouldn't even need window created signal, as `add_window_rules` could be used for this, but I think adding new signal for this is cleaner. As the `add_window_rules` shouldn't be used for this.

Ottatop:
Yea, an open and closed signal is good.

api/lua/pinnacle/window.lua
```diff
@@ -153,13 +153,19 @@ local signal_name_to_SignalName = {
...
+    created = "WindowCreated",
+    destroyed = "WindowDestroyed",
 }
 
 ---@class pinnacle.window.WindowSignal Signals related to compositor events.
 ---@field pointer_enter fun(window: pinnacle.window.WindowHandle)? The pointer entered a window.
 ---@field pointer_leave fun(window: pinnacle.window.WindowHandle)? The pointer left a window.
 ---@field focused fun(window: pinnacle.window.WindowHandle)? The window got keyboard focus.
 ---@field title_changed fun(window: pinnacle.window.WindowHandle, title: string)? A window's title changed.
+---@field layout_mode_changed fun(window: pinnacle.window.WindowHandle, layout_mode: pinnacle.window.v1.LayoutMode)? A window's layout mode changed.
```
Ph4ntomas: The type for `layout_mode` should be `pinnacle.util.v1.LayoutMode`

api/lua/pinnacle/signal.lua
```diff
@@ -97,6 +99,30 @@ local signals = {
         ---@type fun(response: table)
         on_response = nil,
     },
+    WindowLayoutModeChanged = {
+        ---@type grpc_client.h2.Stream?
+        sender = nil,
+        ---@type { callback_id: integer, callback: fun(window: pinnacle.window.WindowHandle, layout_mode: pinnacle.window.v1.LayoutMode) }[]
```
Ph4ntomas: The type for `layout_mode` should be `pinnacle.util.v1.LayoutMode`

Ph4ntomas (commented):
You missed a few types when moving LayoutMode to util, but the lua changes are fine as far as I can tell

api/lua/pinnacle/window.lua
anriha: Thanks, fixed

api/lua/pinnacle/signal.lua
anriha: Thanks, fixed

api/protobuf/pinnacle/util/v1/util.proto
```diff
@@ -37,3 +37,12 @@ enum Dir {
...
 }
+
+enum LayoutMode {
+  LAYOUT_MODE_UNSPECIFIED = 0;
+  LAYOUT_MODE_TILED = 1;
+  LAYOUT_MODE_FLOATING = 2;
+  LAYOUT_MODE_FULLSCREEN = 3;
+  LAYOUT_MODE_MAXIMIZED = 4;
+  LAYOUT_MODE_SPILLED = 5;
+}
```
Ottatop: Moving this enum is a breaking change as it causes v0.1 configs to stop working. Is there a reason this was moved?

api/rust/src/signal.rs
```diff
@@ -294,6 +297,59 @@ signals! {
...
+        /// A window's layout mode changed.
+        ///
+        /// Callbacks receive the window and new layout mode.
+        WindowLayoutModeChanged = {
+            enum_name = LayoutModeChanged,
+            callback_type = Box<dyn FnMut(&WindowHandle, LayoutMode) + Send + 'static>,
+            client_request = window_layout_mode_changed,
+            on_response = |response, callbacks| {
+                let handle = WindowHandle { id: response.window_id };
+                let layout_mode = response.layout_mode().try_into().unwrap_or(LayoutMode::Tiled);
```
Ottatop: Better to just return here if the `try_into` fails

src/state.rs
```diff
@@ -286,6 +286,30 @@ impl State {
...
+                }
+                None => {
+                    // First time seeing this window just set the old state
+                    window.with_state_mut(|state| state.old_layout_mode = Some(current_mode));
+                }
+                _ => {
+                    // No change, do nothing
+                }
+            }
+        }
```
Ottatop: I'd like to extract this into a method on `Pinnacle`

src/handlers/xwayland.rs
```diff
@@ -138,6 +138,8 @@ impl XwmHandler for State {
 
         self.pinnacle.map_window_to(&window, loc);
         self.pinnacle.raise_window(window.clone());
+
+        self.pinnacle.signal_state.window_created.signal(&window);
```
Ottatop: Hm, do we want to trigger this for override-redirect windows? Usually these are used for popups, not true windows.

api/protobuf/pinnacle/util/v1/util.proto
anriha: I thought it makes more sense, as it is now being used in both output.proto and signals.

src/handlers/xwayland.rs
anriha: Honestly not sure, I mainly did it because it is the only other place that creates new windows. But if it is just for xwayland popups I guess it makes sense to not trigger the signal.

anriha: After further though, I decided to remove it.

src/state.rs
anriha: Done, I moved it to `src/window.rs`

api/rust/src/signal.rs
anriha: Done, I did this mainly because that is how the existing api works for `layout_mode_async` but here it doesn't really make sense.

api/protobuf/pinnacle/util/v1/util.proto
Ottatop: Well, I'd rather not have people's configs break when updating, so please move it back.

anriha: Makes sense, I moved it back.

api/lua/pinnacle/signal.lua
Ottatop: Just realized, can you create an `@alias pinnacle.window.LayoutMode` with string variants of the layout mode and use that here? Currently we're passing the raw number to the signal which requires people to import the protobuf definition file.

anriha: I changed it, hopefully correctly. Could you please check it just to be sure? I have basically zero knowledge of lua, so I am basically just trying to replicate existing code. I am removed existing  `enum pinnacle.layout.LayoutMode`, I am fairly certain that it wasn't used anywhere.  And added instead  the `@alias pinnacle.window.LayoutMode`

api/lua/pinnacle/window.lua
```diff
@@ -135,19 +135,27 @@ function window.begin_resize(button)
 end
 
 ---A window's current layout mode.
----@enum (key) pinnacle.layout.LayoutMode
+---@alias pinnacle.window.LayoutMode
+---| "unspecified" The window has unspecified layout mode.
```
Ottatop: I'd remove this "unspecified" variant and just default to "tiled"

anriha: Done

PR #369, opened 2025-10-05
snowcap: add TextInput widget

Ph4ntomas:
The title say it all.

This PR adds TextInput widget, which will be useful as a building block for e.g. prompts.
On top of that, I've enabled support for key-repetition so the prompt themselves feel more natural.

TODO:
[x] - Basic TextInput integration
  [x] - Transmit text in Key Events (very basic for the time being)
  [x] - Enable key repetition
  [x] - on_input event
  [ ] - on_submit event
[ ] - Advanced TextInput integration
  [ ] - Styling support
  [ ] - Copy/Paste support
  [ ] - cursor control from the config ?
  [ ] - Find a way to focus text input from the config
[ ] - Rust configuration support
  [ ] - basic
  [ ] - advanced
[ ] - Lua configuration support
  [ ] - basic
  [ ] - advanced


Ph4ntomas:
> though I strongly recommend doing this in another PR seeing as this one's already up to 2.5k lines

Agree, I'll try to revove the ones introduced by this PR and removable :)

Ph4ntomas:
Allright, I think that's all. 

FYI, the PR is on top of #364. because I needed some of the changes from that PR, and it makes for a cleaner history that cherry-picking the commits I need (plus the eventual rebase hell if I discovered some issue in that PR).

It should rebase cleanly should you choose to first pull the other one, and then rebase + merge this one (which could also be easier to review, I guess ?)

snowcap/api/protobuf/snowcap/widget/v1/widget.proto
```diff
@@ -34,6 +34,36 @@ message Color {
...
+    Linear linear = 1;
+  }
+}
+
+message Background {
+  oneof background {
+    Color color = 1;
+    Gradient gradient = 2;
+  }
+}
```
Ph4ntomas, line 58: I'm having second thought on this.

It matches iced implementation, and will allow future addition of gradient without code duplication, but the current widget have a standalone `background_color`, which make me think the goal might've been to have a `backround_gradient` or a `background_linear` field instead.

I don't feel strongly about which solution ends up implemented, but IMO if this is merged as it is before 0.2 gets out, we should realign other widgets so they use this message for backgrounds and we don't have to duplicate the effort everywhere.

Ph4ntomas: @Ottatop do you have an input on this ? 

Snowcap protobuf v1 was added after after the 0.1, so changing the current way we handle background in container & button would not be a breaking change. However doing so later would.

I can extract this in a standalone PR if you want it.

Ottatop: > do you have an input on this ?

Not at the moment. I'm concerned that we're following the iced API too closely. If they make a breaking change, mirroring it in the protobuf requires a breaking change as well. This is exactly what happened between the v0alpha1 and v1 snowcap protobuf versions. The reason it's such a close mirror boils down to speed of implementation. I'd like to come up with a more detached protobuf API, but that will require a lot of planning.

Ph4ntomas: Yeah, I share your concerns as well.

Okay, then I think I'll drop that next time I rebase. I'd rather keep it consistent with other widgets, and re-add that later.

Ottatop:
> I'd like to come up with a more detached protobuf API

The more I think about this, the less viable it seems without either missing out on some iced functionality or essentially "flattening out" messages. Plus we'd still have to maintain previous messages for a time.

The alternative solution is to follow iced and make additive changes to support any breaking iced updates (like a `TextInput0_16` for example). I'm not *thrilled* about this but it's both much easier to do and gives us the most flexibility in terms of functionality.

I'd like to move forward with adding the background message [here](https://github.com/pinnacle-comp/pinnacle/pull/369#discussion_r2423372568) and deprecating any `background_color` fields. Preferably in another PR.

Ph4ntomas:
> I'd like to move forward with adding the background message https://github.com/pinnacle-comp/pinnacle/pull/369#discussion_r2423372568 and deprecating any background_color fields. Preferably in another PR.

Okay ! Just to clarify, you want another PR to add both the background and the depreciation, right ?

I keep this one as-is, but I cherry-pick back the old commit in a new MR). That way this MR can move along, and depending on which is merged first, the new one will have to remove the background color from text-input, or I'll remove the background_color from this one

Ottatop:
> you want another PR to add both the background and the depreciation, right ?

Yep. I'm probably going to merge it first before reviewing this PR.

snowcap/api/rust/src/decoration.rs
```diff
@@ -187,6 +190,21 @@ impl<Msg> DecorationHandle<Msg> {
...
+    /// [`Operation`]: widget::operation::Operation
+    pub fn operate(&self, operation: widget::operation::Operation) {
+        if let Err(status) = Client::decoration()
+            .operate_decoration(OperateDecorationRequest {
+                decoration_id: self.id.to_inner(),
+                operation: Some(operation.into()),
+            })
+            .block_on_tokio()
+        {
+            error!("Failed to send FocusWidget to {self:?}: {status}");
```
Ottatop: > `FocusWidget`

Think this is a copy-paste error

snowcap/snowcap-api-defs/build.rs
```diff
@@ -20,6 +20,7 @@ fn main() {
 
     tonic_build::configure()
         .file_descriptor_set_path(descriptor_path)
+        .boxed(".snowcap.widget.v1.WidgetDef.widget.text_input")
```
Ottatop, line 23: Just curious, why do we need this?

snowcap/api/rust/src/widget/text_input.rs
```diff
@@ -0,0 +1,701 @@
...
+            && self.icon == other.icon
+            && self.width == other.width
+            && self.padding == other.padding
+            && self.line_height == other.line_height
+            && self.horizontal_alignment == other.horizontal_alignment
+            && self.style == other.style
+            && self.widget_id == other.widget_id
+            && self.callbacks == other.callbacks
+    }
+}
```
Ottatop: Aren't these derivable?

Also realizing that a similar thing is in the mouse area PR and I didn't realize

snowcap/api/rust/src/widget/text_input.rs
```diff
@@ -0,0 +1,701 @@
...
+    /// The [`Color`] to use for the selection's highlight.
+    pub selection: Option<Color>,
+}
+
+impl Style {
+    /// Create a [`Style`] with default values.
+    pub fn new() -> Self {
+        Self {
+            ..Default::default()
+        }
```
Ottatop:
```suggestion
        Default::default()
```

snowcap/api/rust/src/widget/text_input.rs
Ph4ntomas: Yeah, IIRC there was an issue with deriving Debug and PartialEq at some point during development of the two widget (possibly callback related, but I not 100% sure), and I forgot to remove the explicit implementation.

I'll fix the mouse area one in a new PR.

snowcap/api/rust/src/decoration.rs
Ph4ntomas: Kinda, I started with a `focus` function instead of having an `operate` function, and forgot to change the error here.

snowcap/snowcap-api-defs/build.rs
Ph4ntomas: ~~For some reason~~, prost doesn't auto box the text_input, and clippy is not happy with a 1KiB enum
EDIT: Prost only box if it detect a cycle. Since TextInput doesn't have any children, there's no nested WidgetDef, so no implicit boxing.

>    Compiling snowcap-api-defs v0.1.0 (/home/phantomas/sources/pinnacle/pinnacle-next/snowcap/snowcap-api-defs)
   Compiling pinnacle v0.2.2 (/home/phantomas/sources/pinnacle/pinnacle-next)
warning: large size difference between variants
   --> /home/phantomas/sources/pinnacle/pinnacle-next/target/debug/build/snowcap-api-defs-9b5939c64c05c30f/out/snowcap.widget.v1.rs:357:5
    |
357 | /     pub enum Widget {
358 | |         #[prost(message, tag = "2")]
359 | |         Text(super::Text),
    | |         ----------------- the second-largest variant contains at least 128 bytes
360 | |         #[prost(message, tag = "3")]
...   |
377 | |         TextInput(super::TextInput),
    | |         --------------------------- the largest variant contains at least 1032 bytes
378 | |         #[prost(message, tag = "12")]
379 | |         WlrTaskList(::prost::alloc::boxed::Box<super::WlrTaskList>),
380 | |     }
    | |_____^ the entire enum is at least 1032 bytes
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#large_enum_variant
    = note: `#[warn(clippy::large_enum_variant)]` on by default
help: consider boxing the large fields or introducing indirection in some other way to reduce the total size of the enum
    |
377 -         TextInput(super::TextInput),
377 +         TextInput(Box<super::TextInput>),
    |
    
(please disregard the WlrTaskList, it's a WIP POC :) )

snowcap/api/rust/src/widget/text_input.rs
Ph4ntomas: Yeah ... I must've been tired when I wrote that ...

Thanks !

PR #372, opened 2025-10-13
Tag: Add signal to notify config of tag lifetime

### LINKED ISSUE #371: [API] Align Tag signals with ext/workspace-v1
Currently, Tags only gets an active/inactive signal. 

While this allows for a basic bar to be implemented, it means any layer displaying a list of tag must be 'manually' driven on tag addition/removal.

I see two ways of solving that:
- Add support for `ext/workspace-v1` in snowcap, and a widget that would emit corresponding messages.
- Add signal for tag lifetime so the config can properly react to these changes, regardless of snowcap presence.

While both solution would fix my issue in the short term, I think the second one is preferable because the config may need to do some cleanup on tag deletion, so I'll start working on it.

I do think `ext/workspace-v1` support on Snowcap would be nice to have, too

As a side note, maybe at some point the ext/workspace-v1 could be driven by the internal signals (or the other way around).

Ph4ntomas:
This PR adds two signals to tags (one at creation, the other one after removal) so the config can update internal state if needed when signals are added or removed.

Closes #371.

### DIFF
--- a/tests/integration/api/tag.rs
+++ b/tests/integration/api/tag.rs
@@ -1,5 +1,11 @@
+use std::{
+    collections::{HashMap, HashSet},
+    sync::{Arc, Mutex},
+};
+
+use mlua::{UserData, UserDataMethods};
 use pinnacle::{state::WithState, tag::Tag};
-use pinnacle_api::layout::LayoutNode;
+use pinnacle_api::{layout::LayoutNode, signal::TagSignal, tag::TagHandle};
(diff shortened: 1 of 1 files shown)


Ottatop (approved):
Aight lgtm. I mean yea the testing is a little scuffed but if it works it works. Testing in general is kind of all over the place with the whole having to test two whole APIs along with wlcs and the actual compositor, definitely something to figure out. Thank you!

PR #373, opened 2025-10-13
api/lua: Fix XDG_DATA_DIRS usage.

Ph4ntomas:
problem: pinnacle doesn't run properly if XDG_DATA_DIRS is not set.

cause: According to the spec, when this variable is missing or empty /usr/local/share:/usr/share should be used instead, which is not done here.

solution: Add fallback value in lua APIs.


Ottatop (approved):
Thanks!

PR #379, opened 2025-10-21
snowcap: pool surface events on update

Ph4ntomas:
As discussed in #364, here is the standalone PR for snowcap event pooling.

This fix an issue that arose when more than one events are either generated before the config have had the time to generate a new view, or while the view was "in flight"

As an example, if we take a mouse area:
- First view generation. MouseArea has the id 1.
- The user double click on the area, two events are generated: `MsgPress{1}` & `MsgDoubleClick{1}`.
- The config receive `MsgPress{1}`, update itself and generate a new view. MouseArea now has the id 2.
- The config receive `MsgDoubleClick{1}`, the event is discarded because there are no widgets with this id.

This PR does two thing: Pool events so the config gets every pending events at once, and buffer events if we're expecting a new view, meaning the config will always receive all events available and that said events will always be "in sync" with the view.

### DIFF
--- a/snowcap/src/surface.rs
+++ b/snowcap/src/surface.rs
@@ -47,6 +47,7 @@ pub struct SnowcapSurface {
 
     redraw_scheduled: bool,
     pending_view: Option<ViewFn>,
+    waiting_view: bool,
     pub widgets: SnowcapWidgetProgram,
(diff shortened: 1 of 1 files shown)


Ottatop (approved):
Thanks!

Issue #328, opened 2025-07-08
Input Grabbing interface


Ph4ntomas:
TL;DR Adds a way to redirect all input events to the config for processing, instead of only bound inputs.
My main motivation is for modal behaviors, but it could be used for cli/widget inputs

## Rationale

I'm a huge user of modal behavior in software, and not so much a fan of multi key combo.

### Anatomy of a Mode

A Mode is usually started by a keystroke/keycombo, and is quite similar to the concept of 'binding layer' already present. Where they differ is in the presence of arbitrary _action modifiers_ (not to be confounded with _key modifiers_), that can alter the action being performed.

Modes can be thought as an arbitrary string of inputs in the form (this is heavily simplified):
/M?Am*T/ or /M?Tm*A/
M -> enter a given mode (optional if already in that mode)
A/T -> Action & Target
m* -> modifiers. Could be target modifiers, repetition modifiers, etc.

As an example, here are some of the way I usually interact with my WM:
normal mode (act on focus & clients)
/\d*[hjkl]/ -> move focus in direction (hjkl), stepping over \d* clients
/\d*[HLg]/ -> focus tag by direction (HL) stepping over \d* tag, or explicit tag focus when using g
/s\d*[hjkl]/ -> focus Screen in direction (hjkl), stepping over \d+ screen
/m\d+[hlg]/ -> Move client to Tag by direction|globally.

### Existing state

Pinnacle's binding layers only partially address this use-case. While its a good start for basic interactions, you can't encode _action modifiers_ (at least not easily, see Workaround section).

## Enhancement

I'm don't think the compositor should handle all of that, but instead expose an API to request for all input to be redirected to the config/client, upon which we can build things.

While the modal behavior is my primary concern, the same facilities would also enable rudimentary text inputs for (e.g.) a pseudo cli to start processes without having to open a terminal/menu. (as an example, awesomewm has a ctrl+x binding by default).

This would not replace bindings either. The idea would have binding as they exist now, intercepting inputs with the highest priority, then the active grabber, then the windows.

## Workaround

This should already be somewhat achievable through the use of layers. While I haven't tried it yet, you could in theory bind every inputs possible, and process that in a handler for a given layer. 

I'm not sure however that you could cleanly have both keybinds **and** the pseudo grabber at the same time.
The other issue with that workaround is that it adds extra delay

Ottatop:
[...]

PR #383, opened 2025-10-26
Snowcap/Layer: improve key event handling

### LINKED ISSUE #374: [Snowcap] Let captured input go through
Snowcap have an internal subscription on keyboard events, but the events are filtered on the server side.

I would like the ability to handle some key events from the config, regardless of if a widget 'captured' it (the 'captured' flag could be sent so the config can the decide how to process the event.

For context:
I'm using a TextInput for my prompt (from #369).
The default behavior is that the TextInput captures keyboard inputs while focused. 

I'd like to add the following behaviors:
- Close the prompt on Escape => The first Escape is captured by the TextInput, which unfocus itself. The config is not notified of this change.
- Close the prompt on Backspace if the field is empty => The BackSpace is always captured by the TextInput.

Ph4ntomas:
This PR address a few issue I have with the current `Layer::on_key_pressed`
- Add a way to get all events, regardless of whether they are captured (#374).
- Add a way to get `on_release` events (mentioned in #328).
- Add a way to get the `text` associated with the event (mentioned in #328).

I've cherry-picked some commits from #369 (the one to emit the `text` in events).

To avoid breaking current config, `Layer::on_key_pressed` was mostly kept as-is, except I'm explicitly discarding events if they were `captured`, since they used to be discarded on the server side. I don't know if the function should be marked as deprecated, or kept as a short-hand if only key press are needed.

I did not add `Layer::on_key_release`, since: 
- This is already covered by the new `Layer::on_key_event()` function.
- There's no way to have several handlers, and I don't feel like only handling key release events make sense.

closes #374

### DIFF
--- a/snowcap/api/lua/snowcap/layer.lua
+++ b/snowcap/api/lua/snowcap/layer.lua
@@ -164,16 +164,13 @@ function layer.new_widget(args)
     end)
 end
 
----@param on_press fun(mods: snowcap.input.Modifiers, key: snowcap.Key)
-function LayerHandle:on_key_press(on_press)
+---Do something when a key event is received.
+---@param on_event fun(handle: snowcap.layer.LayerHandle, event: snowcap.input.KeyEvent)
+function LayerHandle:on_key_event(on_event)
     local err = client:snowcap_input_v1_InputService_KeyboardKey(
         { id = self.id },
         function(response)
             ---@cast response snowcap.input.v1.KeyboardKeyResponse
-
-            if not response.pressed then
-                return
-            end
 
             local mods = response.modifiers or {}
             mods.shift = mods.shift or false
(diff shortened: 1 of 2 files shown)


snowcap/api/lua/snowcap/layer.lua
```diff
@@ -169,14 +169,48 @@ function layer.new_widget(args)
...
+        end
+    )
+
+    if err then
+        log.error(err)
+    end
+end
+
 ---@param on_press fun(mods: snowcap.input.Modifiers, key: snowcap.Key)
 function LayerHandle:on_key_press(on_press)
```
Ottatop, line 202: This should call `on_key_event` and filter for captures before calling `on_press` so we don't have duplicate code.

snowcap/api/rust/src/layer.rs
```diff
@@ -337,6 +337,35 @@ where
...
+            while let Some(Ok(response)) = stream.next().await {
+                let event = KeyEvent::from(response);
+
+                on_event(handle.clone(), event);
+            }
+        });
+    }
+
     /// Do something on key press.
     pub fn on_key_press(
```
Ottatop, line 379: Same here

PR #387, opened 2025-11-02
snowcap: Allow updating Layers attributes

Ph4ntomas:
This PR adds an `update` function on LayerHandle, to update the Layer attribute.

This is needed for more 'advanced' layer.

As an example, I have a bar with a prompt built with #369. I don't want the bar to have keyboard interactivity, unless a prompt is active.
When the prompt is active, I want all inputs to go to the bar, so that the (now focused) text input can collect inputs.

### DIFF
--- a/snowcap/api/lua/snowcap/layer.lua
+++ b/snowcap/api/lua/snowcap/layer.lua
@@ -14,7 +14,7 @@ local layer_handle = {}
 
 ---@class snowcap.layer.LayerHandle
 ---@field id integer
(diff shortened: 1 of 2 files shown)


Ottatop (approved):
Thanks!

I *will* note that I'm not the biggest fan of specifically `LayerHandle::update` (function signature isn't that ergonomic), but seeing as this is all still an experimental API I'm just gonna merge it. Though expect it to change in the future.

PR #389, opened 2025-11-07
Request view to the server on layer/decoration send_message method

### LINKED ISSUE #388: layer::send_message cause event desync.
### Describe the bug

In #379, we pooled & buffer events sent to the config because the first event would cause a new view to be sent, and any subsequent events were targeting a stale view.

This works well if all events are coming from the server, however, its possible for the config to trigger a re-rendering through the send_message function. The issue is that this new view is not tied to a specific server side-event, and does not trigger the event buffering mechanism, which can effectively de-sync the events, because more than one views can be in-flight. 

### Reproduction steps

Well ... this is awkward.

I don't have a clean way to reproduce that on main. I'm running my own branch (https://github.com/Ph4ntomas/pinnacle/tree/next) & https://github.com/Ph4ntomas/glacier, and this is especially noticeable with the TagList widget & fast scroll events (specifically when using a trackpad, but it's do-able with a mouse).

It's also easier to reproduce using the lua config, since it's a bit slower than the rust one.

### Relevant parts of config

```lua
local function setup_glacier(output)
    Glacier.output(output).prompt = Glacier.widget.prompt()

    Glacier.bar({
        style = {
            background_color = Glacier.color.from_hex('#1a1a1a'),
            first_spacing = -4,
        },
        first = {
            Glacier.widget.taglist({
                output = output,
                style = {
                    spacing = -9,
                },
                inner_view = function(name, style)
                    return Widget.container({
                        child = Widget.row({
                            width = Widget.length.Shrink,
                            spacing = -1,
                            children = {
                                Glacier.separators.arrow.right_inv():to_image(style.bg_color),
                                Glacier.widget.taglist.default_inner_view(name, style),
                                Glacier.separators.arrow.right():to_image(style.bg_color)
                            }
                        }),
                    })
                end
            }),
        },
        output = output
    })
end
```

### Version info

pinnacle 0.1.0

Build Info:
    Build profile: release
    rustc version: 1.90.0
    Branch: makepkg
    Commit: 3e8f235 (Merge remote-tracking branch 'origin/snowcap-layer-update' into next)

### Distro

Arch

### GPU

irrelevant

### Config language

Lua

Ph4ntomas:
This address a small bug where the send_message method could de-sync the server state, leading to events being sent for a stale view.

#379 addressed the issue where multiple events were sent by the server by pooling & buffering them. However, since `send_message` sends a new view after updating the state, it was possible to have views in-flight without the server knowledge, leading to events arriving with stale widget ids if the server sent event in the meantime.

This PR address this de-sync by having the client ask the server to request a new view instead of sending it right after updating its state. 

On server side, the surface is marked with a flag forcing it to send an event, even if there are no message (if there are, no additional event is sent). This forces the server to buffer further events until a new view is received.

On client side, the program state is updated, then the request is sent. If there were no pending event, the client will receive an empty event array, and will not update a second time. If there were pending event, they will be handled normally once the previous view is applied. In that case, the client will receive the messages linked to the server-side event and update normally.

Fixes #388

### DIFF
--- a/snowcap/api/lua/snowcap/grpc/defs.lua
+++ b/snowcap/api/lua/snowcap/grpc/defs.lua
@@ -887,6 +887,11 @@ local snowcap_layer_v1_Layer = {
 
 ---@class snowcap.decoration.v1.UpdateDecorationResponse
 
+---@class snowcap.decoration.v1.ViewRequest
+---@field decoration_id integer?
+
+---@class snowcap.decoration.v1.ViewResponse
+
 ---@class snowcap.input.v0alpha1.Modifiers
 ---@field shift boolean?
 ---@field ctrl boolean?
@@ -1122,6 +1127,8 @@ snowcap.decoration.v1.CloseRequest = {}
 snowcap.decoration.v1.CloseResponse = {}
 snowcap.decoration.v1.UpdateDecorationRequest = {}
 snowcap.decoration.v1.UpdateDecorationResponse = {}
+snowcap.decoration.v1.ViewRequest = {}
+snowcap.decoration.v1.ViewResponse = {}
 snowcap.input = {}
 snowcap.input.v0alpha1 = {}
 snowcap.input.v0alpha1.Modifiers = {}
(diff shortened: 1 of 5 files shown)


Ottatop:
~~Hm it seems this has broken widget updating on v0.2.2 configs, though I'm not sure why that's the case. From what I can see the new server behavior should be the same as the old server behavior on clients before this PR.~~ See below

PR #392, opened 2025-11-27
api: Add option to swap a whole tag to another output in one go.

xMAC94x:
example usage:

```rust
    input::keybind(mod_key | Mod::CTRL, 'l')
        .on_press(move || {
            let Some(monitor1) = output::get_by_name("eDP-1") else {
                return;
            };

            let Some(monitor2) = output::get_by_name("HDMI-1") else {
                return;
            };

            const TAG_NAME: &str = "1";

            if let Some(tag) = tag::get_on_output(TAG_NAME, &monitor1) {
                let result = tag::move_to_output(&monitor2, vec![tag]);
                println!("{result:?}");
            } else {
                let tag = tag::get_on_output(TAG_NAME, &monitor2);
                let result = tag::move_to_output(&monitor1, tag);
                println!("{result:?}");
            }
        })
        .group("Compositor")
        .description("Move Output");
```

### DIFF
--- a/src/api/tag.rs
+++ b/src/api/tag.rs
@@ -181,6 +181,7 @@ where
         .collect::<IndexSet<_>>();
 
     let mut problematic_windows = IndexSet::new();
+    let mut windows_to_update = Vec::new();
 
     for window in state.pinnacle.windows.iter() {
         let (window_id, is_affected_by_move, has_other_output_tag) = window.with_state(|state| {
@@ -190,8 +191,11 @@ where
             (state.id, is_affected_by_move, has_other_output_tag)
         });
 
-        if is_affected_by_move && has_other_output_tag {
-            problematic_windows.insert(window_id);
+        if is_affected_by_move {
+            if has_other_output_tag {
+                problematic_windows.insert(window_id);
+            }
+            windows_to_update.push(window.clone());
         }
     }
 
@@ -221,6 +225,19 @@ where
         if changed {
             state.pinnacle.request_layout(&output);
             state.schedule_render(&output);
+        }
+    }
+
+    if state.pinnacle.space.output_geometry(&new_output).is_some() {
+        let loc = state.pinnacle.floating_loc_for_output(&new_output);
+        for window in windows_to_update {
+            window.with_state_mut(|state| state.set_floating_loc(Some(loc)));
+
+            let layout_mode = window.with_state(|state| state.layout_mode);
+            state.pinnacle.update_window_geometry(
+                &window,
+                layout_mode.is_tiled() || layout_mode.is_spilled(),
+            );
         }
     }
 
--- a/src/window/layout.rs
+++ b/src/window/layout.rs
@@ -226,6 +226,25 @@ impl Pinnacle {
         }
     }
 
+    /// Compute a floating window location for an output.
+    ///
+    /// This is used when a window's tags are moved across outputs to prevent the
+    /// window from "jumping back" to an old floating position.
+    pub(crate) fn floating_loc_for_output(&self, target: &Output) -> Point<i32, Logical> {
+        let output_loc = target.current_location();
+
+        let Rectangle { mut loc, size } = layer_map_for_output(target).non_exclusive_zone();
+
+        // Slightly offset the location so the window is not jammed in a corner
+        let offset = {
+            let (w, h) = size.downscale(100).into();
+            i32::min(w, h)
+        };
+
+        loc += output_loc + Point::new(offset, offset);
+        loc
+    }
+
     /// Move a window to a new output.
     ///
     /// The move itself is done by updating the window tags.
@@ -252,19 +271,7 @@ impl Pinnacle {
 
         window.set_tags_to_output(&target);
 
-        // Reset the floating loc since we're changing output.
-        let output_loc = target.current_location();
-
-        let Rectangle { mut loc, size } = layer_map_for_output(&target).non_exclusive_zone();
-
-        // Slightly offset the location so the window is not jammed in a corner
-        let offset = {
-            let (w, h) = size.downscale(100).into();
-            i32::min(w, h)
-        };
-
-        loc += output_loc + Point::new(offset, offset);
-
+        let loc = self.floating_loc_for_output(&target);
         window.with_state_mut(|state| state.set_floating_loc(Some(loc)));
 
         let layout_mode = window.with_state(|state| state.layout_mode);


api/lua/pinnacle/grpc/defs.lua
```diff
@@ -1360,6 +1360,10 @@ local pinnacle_v1_Backend = {
 ---@field tag_id integer?
 ---@field set_or_toggle pinnacle.util.v1.SetOrToggle?
 
+---@class pinnacle.tag.v1.SwitchOutputRequest
+---@field output_name string?
+---@field tag_names string[]?
```
Ph4ntomas: This should be the tag ids to match protobuf definition, not their names.
You should not edit this file, but run `just build` or `just gen-lua-pb-defs` to generate it instead.

api/lua/pinnacle/tag.lua
```diff
@@ -194,6 +194,37 @@ function tag.remove(tags)
...
+---
+---@param output pinnacle.output.OutputHandle The output to add tags to.
+---@param ... string The names of the new tags.
+---
+---@overload fun(output: pinnacle.output.OutputHandle, tag_names: string[])
+function tag.switch_output(output, ...)
+    local tag_names = { ... }
+    if type(tag_names[1]) == "table" then
+        tag_names = tag_names[1] --[=[@as string[]]=]
+    end
```
Ph4ntomas: The type (L206) should be `TagHandle` here.

I don't think ellipsis are warranted here. Most (if not all) functions in this file use the concrete type in a single parameter (so `TagHandle[]`) instead of ellipsis, and IMO this one shouldn't be different. This will remove the need to fixup the type line 211.

api/lua/pinnacle/tag.lua
```diff
@@ -194,6 +194,37 @@ function tag.remove(tags)
...
+        output_name = output.name,
+        tag_names = tag_names,
+    })
+
+    if err then
+        log.error(err)
+        return {}
+    end
+
+    assert(response)
```
Ph4ntomas: Assert will crash the config if `response` is nil. 

In that context, `nil` could very well be valid too since there's nothing to return if there's no errors. `response` is not returned either way, so you could just discard it.

api/lua/pinnacle/tag.lua
```diff
@@ -194,6 +194,37 @@ function tag.remove(tags)
...
+    end
+
+    local response, err = client:pinnacle_tag_v1_TagService_SwitchOutput({
+        output_name = output.name,
+        tag_names = tag_names,
+    })
+
+    if err then
+        log.error(err)
+        return {}
```
Ph4ntomas: There is no need to return anything here.

api/lua/pinnacle/tag.lua
```diff
@@ -194,6 +194,37 @@ function tag.remove(tags)
...
+function tag.switch_output(output, ...)
+    local tag_names = { ... }
+    if type(tag_names[1]) == "table" then
+        tag_names = tag_names[1] --[=[@as string[]]=]
+    end
+
+    local response, err = client:pinnacle_tag_v1_TagService_SwitchOutput({
+        output_name = output.name,
+        tag_names = tag_names,
+    })
```
Ph4ntomas, line 244: For this part, I you can copy the way tag.remove (the previous function) handles its array of TagHandle and call the client. It should be similar (except `SwitchOutput` has `output_name` as an additional parameter)

Ph4ntomas (changes requested):

Ph4ntomas (commented):
Thanks ! Lua side looks good !

(I can't mark the comments as resolved, feel free to do so)

api/lua/pinnacle/tag.lua
```diff
@@ -194,6 +194,34 @@ function tag.remove(tags)
...
+---
+---#### Example
+---```lua
+---local tag_to_move = Tag.get("1")
+---Tag.switch_output(Output.get_by_name("HDMI-1"), tag_to_move)
+---```
+---
+---@param output pinnacle.output.OutputHandle The output to add tags to.
+---@param tags pinnacle.tag.TagHandle The names of the new tags.
+function tag.switch_output(output, tags)
```
Ottatop: Instead of being a function in this module, I'd like this to be a method on `TagHandle`.

api/rust/src/tag.rs
```diff
@@ -77,6 +77,31 @@ where
...
+/// Move existing tags to the specified output.
+///
+/// # Examples
+///
+/// ```no_run
+/// let output = output::get_by_name("eDP-1")?;
+/// let tag_to_move = tag::get("1")?;
+/// tag::switch_output(&output, tag_to_move);
+/// ```
+pub fn switch_output<I>(output: &OutputHandle, tag_handles: I)
```
Ottatop: Same as lua comment above

api/lua/pinnacle/tag.lua
Ph4ntomas: @Ottatop could we have both ?

IMO the module level function is useful if an output disconnect and you want to retrieve some or all tags.

With the module level function, you can simply call OutputHandle:tags() and forward its return value directly (or after filtering the array)

Ottatop: Hm, I suppose with the module-level function you could move multiple tags at once and windows with all those tags wouldn't have any removed. Yea, we can have both.

api/rust/src/tag.rs
```diff
@@ -316,6 +343,14 @@ impl TagHandle {
             .unwrap();
     }
 
+    /// Move existing tags to the specified output.
+    ///
+    /// See [super::tag::move_to_output] for further information
+    pub fn move_to_output<I>(&self, output: &OutputHandle)
```
Ph4ntomas: Is `<I>` needed here ? I believe it's superfluous

src/api/tag.rs
```diff
@@ -144,3 +149,106 @@ pub fn remove(state: &mut State, tags_to_remove: Vec<Tag>) {
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            TagMoveToOutputError::OutputDoesNotExist(output) => write!(
+                f,
+                "Tried to move tags to output {} but it doesn't exist",
+                output.0
+            ),
+            TagMoveToOutputError::SameWindowOnTwoOutputs(tags) => write!(
+                f,
+                "executing this operation would result in at least one window be in tags {:?} which run on different outputs. This is forbidden.",
```
cassandracomar:
```suggestion
                "executing this operation would put the same window in tags {:?} on two separate outputs at once. This is forbidden.",
```

xMAC94x: adjusted but made it plural, because it might result in [1,2,3,4] where window A is in 1 and 2. window B is in 1 and 3 and window C ist in 1 and 4

src/api/tag.rs
```diff
@@ -144,3 +149,106 @@ pub fn remove(state: &mut State, tags_to_remove: Vec<Tag>) {
...
+pub fn move_to_output<T>(
+    state: &mut State,
+    tags_to_move: T,
+    output_name: OutputName,
+) -> Result<(), TagMoveToOutputError>
+where
+    T: IntoIterator<Item = Tag>,
+{
+    let Some(new_output) = output_name.output(&state.pinnacle) else {
+        return Err(TagMoveToOutputError::OutputDoesNotExist(output_name));
```
Ottatop: Why do we need to return the nonexistent output? The client already knows what output is passed in.

src/api/tag.rs
```diff
@@ -144,3 +149,106 @@ pub fn remove(state: &mut State, tags_to_remove: Vec<Tag>) {
...
+                let contains = tags_to_move.contains(tag);
+                affected = contains || affected;
+                contains_other_tags = !contains || contains_other_tags;
+
+                if affected && contains_other_tags {
+                    to_evaluate_tags.insert(tag.clone());
+                }
+            }
+        })
+    }
```
Ottatop: I'm having a little trouble wrapping my head around this. From what I'm seeing, this will, for all windows, add all tags at and after the first found tag that's in the move set to the evaluation set. Is that right?

src/api/tag.rs
```diff
@@ -144,3 +149,106 @@ pub fn remove(state: &mut State, tags_to_remove: Vec<Tag>) {
...
+        if output.name() != output_name.0 {
+            output.with_state(|state| {
+                for tag in to_evaluate_tags
+                    .extract_if(.., |to_evalaute_tag| state.tags.contains(to_evalaute_tag))
+                {
+                    tags_on_other_output.push(tag);
+                }
+            })
+        }
+    }
```
Ottatop: I'm gonna be honest I have no idea what this is supposed to do or what the intent is

src/api/tag.rs
```diff
@@ -144,3 +149,106 @@ pub fn remove(state: &mut State, tags_to_remove: Vec<Tag>) {
...
+
+    if !tags_on_other_output.is_empty() {
+        return Err(TagMoveToOutputError::SameWindowOnTwoOutputs(
+            tags_on_other_output,
+        ));
```
Ottatop, line 205: Going by the name of this variable, how useful is returning the tags on other outputs (whatever that is)? I feel like it would be much more useful to return the windows that would be on multiple outputs so you can move them if needed.

api/lua/pinnacle/tag.lua
```diff
@@ -194,6 +194,40 @@ function tag.remove(tags)
...
+
+    local response, err = client:pinnacle_tag_v1_TagService_MoveToOutput({
+        output_name = output.name,
+        tag_ids = ids,
+    })
+
+    if err then
+        log.error(err)
+        return false, response
+    end
```
Ottatop, line 249: `err` occurs when there are gRPC status errors ([tonic docs](https://docs.rs/tonic/latest/tonic/struct.Status.html)) or http/io errors. In this case, we're returning a possible error through the response type itself, so you need to instead check if `response.output_does_not_exist ~= nil` and if `response.same_window_on_two_outputs ~= nil` to determine if an error happened.

api/protobuf/pinnacle/tag/v1/tag.proto
```diff
@@ -22,6 +22,21 @@ message RemoveRequest {
...
+message MoveToOutputRequest {
+    string output_name = 1;
+    repeated uint32 tag_ids = 2;
+}
+// oneof has a implicit None, if kind == None -> successful execution
+message MoveToOutputResponse {
+    oneof kind {
+      string output_does_not_exist = 2;
+      MoveToOutputSameWindowOnTwoOutputs same_window_on_two_outputs = 3;
+    }
```
Ottatop: Can we extract `kind` into its own `Error` message nested in here?

api/protobuf/pinnacle/tag/v1/tag.proto
```diff
@@ -22,6 +22,21 @@ message RemoveRequest {
...
+}
+message MoveToOutputSameWindowOnTwoOutputs {
+    repeated uint32 tag_ids = 1;
+}
```
Ottatop: I'd like to nest this under the `Error` type above when you add it.

src/api/tag.rs
Ph4ntomas: If I understand correctly, the goal here would be to take all tags that are not in `tags_to_move`, and add them to the `to_evaluate_tags`.

As Ottatop mentioned, the current function is a bit weird, because it will ignore tags until one tag to move is found.

`IndexSet::difference` should already do what you want (return all tags in `state.tags` that are not in `tags_to_move`).

src/api/tag.rs
Ph4ntomas: I think I understand the goal here, but I'm not sure it's needed.

The code essentially goes over every output (except the target), and compute the intersection between the output tags and `to_evaluate_tags`.
If there is at least one tag on another output, the tag move will be aborted, because at least one window would end up on two outputs.

If that's the case, I think this is somewhat overkill. This code assumes that a window could be on multiple output before the tag move (because one of the non-moved tag could be on our target output), only to then refuse to perform the operation because the window would end up on multiple outputs.

If the goal here is to prevent the "window on multiple output" state, it would be best to assume the state is correct before the call, and simplify the check (if a window is on multiple tags, all tags have to move for the move to occur (which would translate to `to_evaluate_tags.empty()`)).

I know it's currently possible set tags from multiple output, but that should be fixed in the function that can cause this state to appears to begin with.

src/api/tag.rs
xMAC94x: We don't need to, but its a common pattern sometimes beeing used https://doc.rust-lang.org/src/alloc/boxed/convert.rs.html#313 . I can also remove it if you want.

src/api/tag.rs
xMAC94x: Yeah, it would prob be more useful, I don't remember why I choose to return tags instead, either windows was to complicated or one sometimes you use the tags to bringt it to work. 

Maybe both is also helpful, the window and the conflicting tag. Or we leave finding the conflicting tag to the caller of the function

src/api/tag.rs
Ottatop: Yea go ahead and remove it.

> its a common pattern sometimes beeing used https://doc.rust-lang.org/src/alloc/boxed/convert.rs.html#313 .

That pattern is usually reserved for fallible operations that take ownership of the operand to allow callers to recover the value on failure. In our case, the Rust API is passing in `&OutputHandle` and does not need to recover the handle on failure because ownership was never given up, and the Lua API merely passes in a shallow copy of the `OutputHandle` table.

Even if we wanted to do this, we don't need to pass the handle through to the server and back because the client has the handle and can return it without the roundtrip.

src/api/tag.rs
Ottatop: If you have the problem windows then you should have all the information you need from what tags you're moving and the tags on those windows, no? I just don't see the utility in passing in extra tag information.

src/api/tag.rs
cassandracomar: the wording is just awkward. let's make error messages as clear as possible.

api/lua/pinnacle/tag.lua
xMAC94x: that the user of this APIs tasks. as we forward the response.

api/protobuf/pinnacle/tag/v1/tag.proto
xMAC94x: can you give me an example .proto syntax how you like it ?

Ottatop:
```proto
message MoveToOutputResponse {
    message Error {
        message SameWindowOnTwoOutputs {
            repeated uint32 window_ids = 1;
        }
        
        oneof kind {
            google.protobuf.Empty output_does_not_exist = 1;
            SameWindowOnTwoOutputs same_window_on_two_outputs = 2;
        }
    }

    optional Error error = 1;
}
```

api/lua/pinnacle/tag.lua
Ottatop: I meant you need to also check that those two are not nil to return `true`. As it stands you return `true` regardless of whether there's an `output_does_not_exist` or `same_window_on_two_outputs` error.

api/lua/pinnacle/tag.lua
```diff
@@ -194,6 +194,40 @@ function tag.remove(tags)
...
 
+---Move existing tags to the specified output.
+---
+---#### Example
+---```lua
+---local tag_to_move = Tag.get("1")
+---Tag.move_to_output(Output.get_by_name("HDMI-1"), tag_to_move)
+---```
+---
+---@param output pinnacle.output.OutputHandle The output to add tags to.
```
Ottatop:
```suggestion
---@param output pinnacle.output.OutputHandle The output to move tags to.
```

api/lua/pinnacle/tag.lua
```diff
@@ -194,6 +194,40 @@ function tag.remove(tags)
...
+---@param tags pinnacle.tag.TagHandle The names of the new tags.
```
Ottatop:
```suggestion
---@param tags pinnacle.tag.TagHandle[] The tags to move.
```

api/lua/pinnacle/tag.lua
```diff
@@ -194,6 +194,40 @@ function tag.remove(tags)
...
+---
+---@return boolean ok `true` on success.
+---@return pinnacle.tag.v1.MoveToOutputResponse|nil err Error on failure.
```
Ottatop: Can you create a custom error class for this? Generated types should not be exposed to the user. Something like
```lua
---@class pinnacle.tag.MoveToOutputError
---`true` if the output does not exist.
---@field output_does_not_exist boolean?
---This operation would cause the provided windows to be on multiple outputs.
---@field same_window_on_two_outputs pinnacle.window.WindowHandle[]?
```
You'll need to wrap the returned ids in a window handle with `require("pinnacle.window").handle.new_from_table(window_ids)`.

api/lua/pinnacle/tag.lua
```diff
@@ -194,6 +194,40 @@ function tag.remove(tags)
     end
 end
 
+---Move existing tags to the specified output.
```
Ottatop:
```suggestion
---Moves existing tags to the specified output.
```

api/lua/pinnacle/tag.lua
```diff
@@ -194,6 +194,40 @@ function tag.remove(tags)
...
+---
+---#### Example
+---```lua
+---local tag_to_move = Tag.get("1")
+---Tag.move_to_output(Output.get_by_name("HDMI-1"), tag_to_move)
```
Ottatop:
```suggestion
---Tag.move_to_output(Output.get_by_name("HDMI-1"), { tag_to_move })
```

api/lua/pinnacle/tag.lua
```diff
@@ -322,6 +356,31 @@ function TagHandle:toggle_active()
     end
 end
 
+---Move existing tags to the specified output.
```
Ottatop:
```suggestion
---Moves existing tags to the specified output.
```

api/lua/pinnacle/tag.lua
```diff
@@ -322,6 +356,31 @@ function TagHandle:toggle_active()
...
+---
+---@see pinnacle.tag.move_to_output - for further information
+---@param output pinnacle.output.OutputHandle The output to add tags to.
```
Ottatop:
```suggestion
---@param output pinnacle.output.OutputHandle The output to move this tag to.
```

api/rust/src/tag.rs
```diff
@@ -181,6 +182,63 @@ pub fn remove(tags: impl IntoIterator<Item = TagHandle>) {
...
+pub enum TagMoveToOutputError {
+    /// The requested output to move the tag to, does not exist
+    OutputDoesNotExist,
+
+    /// Moving the Tag to another output would result in having the same window in multiple tags.
+    /// It contains a list of windows that would be on multiple outputs.
+    SameWindowOnTwoOutputs(Vec<WindowHandle>),
+}
+
+/// Move existing tags to the specified output.
```
Ottatop:
```suggestion
/// Moves existing tags to the specified output.
```

api/rust/src/tag.rs
```diff
@@ -316,6 +374,13 @@ impl TagHandle {
             .unwrap();
     }
 
+    /// Move existing tags to the specified output.
```
Ottatop:
```suggestion
    /// Moves this tag to the specified output.
```

src/api/tag.rs
```diff
@@ -144,3 +149,95 @@ pub fn remove(state: &mut State, tags_to_remove: Vec<Tag>) {
...
+                write!(f, "Tried to move tags to output but it doesn't exist")
+            }
+            TagMoveToOutputError::SameWindowOnTwoOutputs(window_ids) => write!(
+                f,
+                "executing this operation would put the same windows {:?} on two separate outputs at once. This is forbidden.",
+                window_ids.iter().map(|id| id.0).collect::<Vec<_>>()
+            ),
+        }
+    }
+}
```
Ottatop: Is this `Display` impl used anywhere? Given that we *are* the compositor I don't see why we would need this.

src/api/tag.rs
```diff
@@ -144,3 +149,95 @@ pub fn remove(state: &mut State, tags_to_remove: Vec<Tag>) {
...
+        .collect::<IndexSet<_>>();
+
+    let mut windows_on_other_output = IndexSet::new();
+
+    for window in state.pinnacle.windows.iter() {
+        let (window_id, has_other_output_tag) = window.with_state(|state| {
+            let has_other_output_tag = state
+                .tags
+                .iter()
+                .any(|tag| tags_on_other_outputs.contains(tag));
```
Ottatop:
```suggestion
            let has_other_output_tag = !state.tags.is_disjoint(&tags_on_other_outputs);
```

api/rust/src/tag.rs
```diff
@@ -181,6 +182,63 @@ pub fn remove(tags: impl IntoIterator<Item = TagHandle>) {
         .unwrap();
 }
 
+/// Error that happends during moving a tag to a different output.
```
Ottatop:
```suggestion
/// Error that happens when moving tags to a different output.
```

api/rust/src/tag.rs
```diff
@@ -181,6 +182,63 @@ pub fn remove(tags: impl IntoIterator<Item = TagHandle>) {
...
+#[derive(Debug, PartialEq, Clone)]
+pub enum TagMoveToOutputError {
```
Ottatop: Could probably rename this to just `MoveToOutputError`

api/rust/src/tag.rs
```diff
@@ -316,6 +374,13 @@ impl TagHandle {
             .unwrap();
     }
 
+    /// Move existing tags to the specified output.
+    ///
+    /// See [super::tag::move_to_output] for further information
```
Ottatop:
```suggestion
    /// See [tag::move_to_output][crate::tag::move_to_output] for more information.
```

src/api/tag.rs
xMAC94x: Its necessary for a https://doc.rust-lang.org/std/error/trait.Error.html to implement Display. And its prob good measure to implement `std::error::Error` for our error enums or ?

src/api/tag.rs
xMAC94x: rewrote this to return windows and heavily rely on `IndexSet::is_disjoint`

api/lua/pinnacle/tag.lua
```diff
@@ -135,7 +135,7 @@ end
 ---local tags = Tag.add(Output.get_by_name("HDMI-1"), tag_names)
 ---```
 ---
----@param output pinnacle.output.OutputHandle The output to add tags to.
+---@param output pinnacle.output.OutputHandle The output to move this tag to.
```
Ottatop: Wait, did I leave a review comment in the wrong place? Or did you change the wrong thing? This shouldn't be changed, this is for `tag.add`.

api/lua/pinnacle/tag.lua
```diff
@@ -205,6 +244,35 @@ local signal_name_to_SignalName = {
...
 
+---@class pinnacle.tag.MoveToOutputError
+---`true` if the output does not exist.
+---@field output_does_not_exist boolean?
+---This operation would cause the provided windows to be on multiple outputs.
+---@field same_window_on_two_outputs pinnacle.window.WindowHandle[]?
+
+---@param response pinnacle.tag.v1.MoveToOutputResponse?
+---@return pinnacle.tag.MoveToOutputError|nil
+local function move_to_output_error_from_response(response)
```
Ottatop: This needs to be defined above where you use it because it's a local.

api/lua/pinnacle/tag.lua
```diff
@@ -322,6 +390,36 @@ function TagHandle:toggle_active()
     end
 end
 
+---Moves existing tags to the specified output.
```
Ottatop:
```suggestion
---Moves this tag to the specified output.
```

api/lua/pinnacle/tag.lua
```diff
@@ -322,6 +390,36 @@ function TagHandle:toggle_active()
...
+---
+---@see pinnacle.tag.move_to_output - for further information
+---@param output pinnacle.output.OutputHandle The output to move tags to.
```
Ottatop:
```suggestion
---@param output pinnacle.output.OutputHandle The output to move this tag to.
```

api/lua/pinnacle/tag.lua
```diff
@@ -322,6 +390,36 @@ function TagHandle:toggle_active()
...
+        log.error(err)
+        return false, nil
+    end
+
+    local move_error = move_to_output_error_from_response(response)
+    if move_error then
+        return false, move_error
+    end
+
+    return true, nil
```
Ottatop: You can replace this entire body with `return tag.move_to_output(output, { self })`.

api/protobuf/pinnacle/tag/v1/tag.proto
```diff
@@ -22,6 +22,26 @@ message RemoveRequest {
     repeated uint32 tag_ids = 1;
 }
 
+message MoveToOutputRequest {
+    string output_name = 1;
+    repeated uint32 tag_ids = 2;
+}
+// oneof has a implicit None, if kind == None -> successful execution
```
Ottatop:
```suggestion

```

src/api/tag.rs
Ottatop: > Its necessary for a https://doc.rust-lang.org/std/error/trait.Error.html to implement Display.

But this doesn't implement `Error`.

> And its prob good measure to implement `std::error::Error` for our error enums or ?

Well given we're only using the type to convert to the protobuf type I don't see why we would need it to impl `Error`.

src/api/tag.rs
```diff
@@ -144,3 +150,95 @@ pub fn remove(state: &mut State, tags_to_remove: Vec<Tag>) {
...
+        }
+
+        if changed {
+            state.pinnacle.request_layout(&output);
+            state.schedule_render(&output);
+        }
+    }
+
+    state.pinnacle.update_xwayland_stacking_order();
+    ext_workspace::refresh(state);
```
Ottatop: I don't think we need this here, it's already called at the end of every event loop cycle and I'd rather not sprinkle this in multiple places like a bunch of other things.

api/lua/pinnacle/tag.lua
xMAC94x: I adjusted the wrong member function. sorry.

api/lua/pinnacle/tag.lua
xMAC94x: thank you, wasnt sure how to convert self to `tags`. implemented.

src/api/tag.rs
xMAC94x: huh, interestingly. I forgot to add `Error` trait for it. I removed it for now and added doc-comments instead. Hope thats okay. stuff like `std::error::Error` might be reevaluated once there are more error types.

I somehow tested wrong, I removed `ext_workspace::refresh(state)` again and it still works. sorry for the confusion.

Pushed all changes :)

xMAC94x:
> Ok I've pushed a test for `Tag::move_to_output`. If you run `just install test tag_move_to_output` ...

Thank you for adding a test, I can see the failure.

However I couldn't find a fix yet. I am also lacking a few information:
 - are floating windows assigned any tags at all ? are they assigned any outputs ?
 - what is our expected behavior for floating windows ? Do we want them to move to the new location of the tag ?
 - can a floating window be on multiple tags at once ?
 - i just tested that fullscreen windows also dont move. Should a fullscreen window be more bound the output ? or tag ? (my guess would be: the difference between fullscreen and maximized is: fullscreen = glued to the output, maximized = glued to the tag? )

api/lua/pinnacle/tag.lua
```diff
@@ -135,7 +135,7 @@ end
 ---local tags = Tag.add(Output.get_by_name("HDMI-1"), tag_names)
 ---```
 ---
----@param output pinnacle.output.OutputHandle The output to add tags to.
+---@param output pinnacle.output.OutputHandle The output to add this tag to.
```
Ottatop: This still isn't the same as it was before
```suggestion
---@param output pinnacle.output.OutputHandle The output to add tags to.
```

Ottatop (changes requested):
Other than this lgtm

api/lua/pinnacle/tag.lua
xMAC94x: @Ottatop whoopsie. Pushed the fix

Ottatop (approved):
Thank you!

Issue #386, opened 2025-10-30
layout requests are no longer automatic


cassandracomar:
### Describe the bug

layout requests are no longer sent automatically after common operations (non-exhaustive, just what I've noticed):
1. toggling/switching tags
2. after window creation

they've never been sent after e.g. swapping windows within a layout.

this especially affects GTK applications as they're just not rendered after creation. manually requesting layout from the config cleanly works around the issue.

this workaround is probably fine -- ensuring that the user config requests layout after window layout needs to change isn't onerous. however, the need to do this isn't documented and it's not part of the default config we ship. we should either ensure that a layout request is sent uniformly after all operations that can affect displayed windows or we should update the default config and documentation so users know to request layout after such operations themselves.

### Version info

commit: 4cc7283046f474450202cbdad1deacfc79392bc8

### Distro

NixOS

### Config language

Rust

Ph4ntomas:
> this especially affects GTK applications as they're just not rendered after creation. manually requesting layout from the config cleanly works around the issue.

I don't observe the same behavior on my end. Can you reproduce the issue with one of the default configs ?

cassandracomar:
sorry, got sick this past month. I haven't had a chance to do a full bisection but I think this is related to server-side decorations. you can see my full, [working config with latest master](https://github.com/cassandracomar/pinnacle-config/blob/a1cf6b98fe85a239620fa1778f9762f4af0df832/src/main.rs). if I swap back to client-side decorations, I no longer need to issue layout requests or re-decorate windows after each update to window geometry.

Issue #385, opened 2025-10-30
fullscreening/maximizing windows doesn't update window geometry


cassandracomar:
### Describe the bug

maximizing/fullscreening a titled window will leave the gaps visible until deactivating and reactivating the tag. I've tried requesting layout explicitly both with the `LayoutModeChanged` signal and on the keybind itself but the only way to get the window geometry to update appears to be to deactivate the tag (or otherwise hide the surface) and then reactivate it.

weirdly, if there are multiple windows open, the non-maximized/fullscreened window(s) will have their layout updated to not include the maximized/fullscreened window. however, the maximized/fullscreened window retains its original geometry until the surface is hidden and re-exposed.

I think this is related to the layout mode work but not positive. I've generally had to go through my config and stick in manual layout requests because layout isn't automatically re-requested any longer after e.g. toggling a tag, creating a window, et cetera (I'll open a separate issue for this).

### Reproduction steps

starting with a tiled window -- it's easier to see this if your tiled layout includes gaps:
1. maximize the window
2. deactivate the tag
3. reactivate the tag
4. note that it's only now that the window is actually maximized

repeat for fullscreened windows.

### Relevant parts of config

```rust
    window::connect_signal(WindowSignal::LayoutModeChanged(Box::new({
        let requester = layout_requester.clone();
        move |_win, _layout_mode| {
            requester.request_layout();
        }
    })));

    input::keybind(mod_key, 'f')
        .on_press({
            let requester = layout_requester.clone();
            move || {
                if let Some(window) = window::get_focused() {
                    window.toggle_fullscreen();
                    window.raise();
                    requester.request_layout();
                }
            }
        })
        .group("Window")
        .description("Toggle fullscreen on the focused window");

    // `mod_key + m` toggles maximized
    input::keybind(mod_key, 'm')
        .on_press({
            let requester = layout_requester.clone();
            move || {
                if let Some(window) = window::get_focused() {
                    window.toggle_maximized();
                    window.raise();
                    requester.request_layout();
                }
            }
        })
        .group("Window")
        .description("Toggle maximized on the focused window");
```

### Version info

commit: 4cc7283046f474450202cbdad1deacfc79392bc8

### Distro

NixOS

### Config language

Rust

Ph4ntomas:
Same as #386, I can't reproduce with the default config, could you try with either default config on your end ?

PR #401, opened 2025-12-13
nix: disable wlcs for the final package

anriha:
The `wlcs` feature is currently enabled in the final nix package. Which doesn't make sense for the final user package. This removes `wlcs` for the build.

### DIFF
--- a/nix/packages/default.nix
+++ b/nix/packages/default.nix
@@ -16,8 +16,6 @@
   libdisplay-info,
   git,
   libgbm,
-  writeScriptBin,
-  wlcs,
   rustc,
   cargo,
   makeWrapper,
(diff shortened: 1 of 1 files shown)


Ottatop (approved):
Well this seems fine and I don't know any better so we're just gonna hit the merge button

anriha:
If it helps I have been running this since opening this PR and it works great. And it has fixed some big issues with layout logic. As without this, it doesn't work properly, mainly thanks to behavior like in here: https://github.com/pinnacle-comp/pinnacle/commit/0e7cbd68aaf46041196141c3da5d1ced981cad65 as it behaves differently with wlcs feature.

Ph4ntomas:
@anriha when you mention layout logic, do you mean something like the issues raised in #385 and #386 ?

Ottatop:
> And it has fixed some big issues with layout logic

*Oh*, this was enabling the wlcs feature on non-test builds, wasn't it?

PR #402, opened 2025-12-14
[Snowcap] Implement support for Popup

### LINKED ISSUE #400: [Snowcap] Support for popup surface
Hello,

I was working on an implementation of SNI/DBusMenu/Notification for my config/config lib.
Currently, support for DBusMenu can't be done cleanly because there's no support for Popups.

My plan was to add a new type, `SnowcapPopup`, which would hold a `SnowcapSurface`. The idea would be to make such type similar to `SnowcapLayer` and `SnowcapDecoration`. 
The `SnowcapPopup` could be created from a `SnowcapLayer`, a `SnowcapDecoration`, or another `SnowcapPopup`. The parent object need to own the child object so they are discarded from the top-down.

Other consideration can be added as development progress. One thing I'm not yet sure how to handle is the popup grab behavior, since that need to be call in reaction to a parent's wayland event.

@Ottatop are you ok with the (currently rough) idea ? Let me know if you're ok with me trying to implement this.

Ph4ntomas:
~~Until I rebase everything, I'm basing this on my own branch, since this allows me to work on this without breaking my config.~~ Mostly rebased. I removed commits from open PR this one doesn't depends upon.

Interesting stuff starts here: https://github.com/pinnacle-comp/pinnacle/pull/402/commits/9b7332a2ace9cc976ab7ebcce7bae31200190350

TODO:
- [x] Add `SnowcapPopup`
- [x] Renders widgets
- [x] Manage popup parents
  - [x] Layer
  - [x] Popup
  - [x] ~~Decoration~~ Not planned for this MR, since the protocol doesn't support creating popups.
- [x] Correct closing order
- [x] Position popup
  - [x] At pointer
  - [x] Relative to element
  - [x] Offset
  - [x] Gravity
  - [x] Anchor
- [x] PopupGrab / KeyboardInput
  - [x] Basic implementation
  - [x] Make it possible to opt out from the keyboard grab
- [x] Only allow one child popup
- [x] Constraint adjustment
- [x] Popup repositioning ?
- [x] Rust API
- [x] Support Operation
- [x] Document all the thing
- [x] Pinnacle: Fix support for reactive popup

https://github.com/pinnacle-comp/pinnacle/pull/402/commits/8a79256215006ccaf20c8bb861c447457bf9749d could be squashed with the one introducing operation (from #369). The old implementation bring nothing to the table and is somewhat broken anyway.

closes #400

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4489,6 +4489,7 @@ dependencies = [
 name = "snowcap-api"
 version = "0.1.0"
 dependencies = [
+ "bitflags 2.10.0",
  "from_variants",
  "futures",
  "hyper-util",
--- a/snowcap/api/rust/Cargo.toml
+++ b/snowcap/api/rust/Cargo.toml
@@ -5,6 +5,7 @@ edition = "2024"
 rust-version.workspace = true
 
 [dependencies]
+bitflags = { workspace = true }
 from_variants = "1.0.2"
 futures = { workspace = true }
 hyper-util = { workspace = true }
--- a/snowcap/api/rust/src/client.rs
+++ b/snowcap/api/rust/src/client.rs
@@ -6,6 +6,7 @@ use snowcap_api_defs::snowcap::{
     decoration::v1::decoration_service_client::DecorationServiceClient,
     input::v1::input_service_client::InputServiceClient,
     layer::v1::layer_service_client::LayerServiceClient,
+    popup::v1::popup_service_client::PopupServiceClient,
     widget::v1::widget_service_client::WidgetServiceClient,
 };
 use tokio::sync::{RwLock, RwLockReadGuard};
@@ -20,6 +21,7 @@ pub struct Client {
     input: InputServiceClient<Channel>,
     widget: WidgetServiceClient<Channel>,
     decoration: DecorationServiceClient<Channel>,
+    popup: PopupServiceClient<Channel>,
 }
 
 impl Client {
@@ -37,6 +39,10 @@ impl Client {
 
     pub fn layer() -> LayerServiceClient<Channel> {
         Self::get().layer.clone()
+    }
+
+    pub fn popup() -> PopupServiceClient<Channel> {
+        Self::get().popup.clone()
     }
 
     pub fn input() -> InputServiceClient<Channel> {
@@ -57,6 +63,7 @@ impl Client {
             input: InputServiceClient::new(channel.clone()),
             widget: WidgetServiceClient::new(channel.clone()),
             decoration: DecorationServiceClient::new(channel.clone()),
+            popup: PopupServiceClient::new(channel.clone()),
         }
     }
 }
--- a/snowcap/api/rust/src/layer.rs
+++ b/snowcap/api/rust/src/layer.rs
@@ -19,6 +19,7 @@ use crate::{
     BlockOnTokio,
     client::Client,
     input::{KeyEvent, Modifiers},
+    popup::{self, AsParent},
     widget::{self, Program, WidgetDef, WidgetId, WidgetMessage},
 };
 
@@ -383,5 +384,11 @@ where
 
             on_press(handle, event.key, event.mods)
         });
+    }
+}
+
+impl<Msg> AsParent for LayerHandle<Msg> {
+    fn as_parent(&self) -> crate::popup::Parent {
+        popup::Parent(popup::ParentInner::Layer(self.id))
     }
 }
--- a/snowcap/api/rust/src/lib.rs
+++ b/snowcap/api/rust/src/lib.rs
@@ -15,6 +15,7 @@ mod client;
 pub mod decoration;
 pub mod input;
 pub mod layer;
+pub mod popup;
 pub mod widget;
 
 use client::Client;
--- a/snowcap/api/rust/src/popup.rs
+++ b/snowcap/api/rust/src/popup.rs
@@ -1,0 +1,551 @@
+//! Support for popup surface widgets using `xdg-shell::xdg_popup`
+
+use std::collections::HashMap;
+
+use bitflags::bitflags;
+use snowcap_api_defs::snowcap::{
+    input::v1::{KeyboardKeyRequest, keyboard_key_request::Target},
+    popup::{
+        self,
+        v1::{CloseRequest, NewPopupRequest, OperatePopupRequest, UpdatePopupRequest, ViewRequest},
+    },
+    widget::v1::{GetWidgetEventsRequest, get_widget_events_request},
+};
+use tokio::sync::mpsc::UnboundedSender;
+use tokio_stream::StreamExt;
+use xkbcommon::xkb::Keysym;
+
+use crate::{
+    BlockOnTokio,
+    client::Client,
+    input::{KeyEvent, Modifiers},
+    widget::{self, Program, WidgetDef, WidgetId, WidgetMessage},
+};
+
+#[derive(Clone, Copy, PartialEq, Eq, Hash)]
+pub(crate) enum ParentInner {
+    Layer(WidgetId),
+    Popup(WidgetId),
+}
+
+/// Popup Parent surface.
+#[derive(Clone, Copy, PartialEq, Eq, Hash)]
+pub struct Parent(pub(crate) ParentInner);
+
+/// Position the Popup will be placed at.
+#[allow(missing_docs)]
+#[derive(Debug, Clone, PartialEq)]
+pub enum Position {
+    /// Position the anchor point at the cursor.
+    AtCursor,
+    /// Position the anchor at an arbitrary point.
+    Point { x: f32, y: f32 },
+    /// Position the anchor on a Rectangle boundaries.
+    Rectangle {
+        x: f32,
+        y: f32,
+        width: f32,
+        height: f32,
+    },
+    /// Position the anchor on a Widget boundaries.
+    AtWidget(String),
+}
+
+impl Position {
+    /// Create a new Position to place a Popup at the cursor location.
+    pub fn at_cursor() -> Self {
+        Position::AtCursor
+    }
+
+    /// Create a new Position to place a Popup at an arbitrary point.
+    pub fn point(x: f32, y: f32) -> Self {
+        Position::Point { x, y }
+    }
+
+    /// Create a new Position to place a Popup on an arbitrary rectangle.
+    pub fn rectangle(x: f32, y: f32, width: f32, height: f32) -> Self {
+        Position::Rectangle {
+            x,
+            y,
+            width,
+            height,
+        }
+    }
+
+    /// Create a new Position to place a Popup relative to a Widget.
+    pub fn at_widget(id: impl Into<String>) -> Self {
+        Position::AtWidget(id.into())
+    }
+}
+
+/// Position of the anchor point on the anchor rectangle.
+#[allow(missing_docs)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
+pub enum Anchor {
+    None,
+    Top,
+    Bottom,
+    Left,
+    Right,
+    TopLeft,
+    TopRight,
+    BottomLeft,
+    BottomRight,
+}
+
+/// Direction of the gravity of the Popup.
+#[allow(missing_docs)]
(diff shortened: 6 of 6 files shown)


snowcap/api/protobuf/snowcap/popup/v1/popup.proto
```diff
@@ -0,0 +1,85 @@
...
+  GRAVITY_TOP = 1;
+  GRAVITY_BOTTOM = 2;
+  GRAVITY_LEFT = 3;
+  GRAVITY_RIGHT = 4;
+  GRAVITY_TOP_LEFT = 5;
+  GRAVITY_TOP_RIGHT = 6;
+  GRAVITY_BOTTOM_LEFT = 7;
+  GRAVITY_BOTTOM_RIGHT = 8;
+  GRAVITY_NONE = 9;
+}
```
Ph4ntomas: I'm having some doubt on this.

It make sense to have different enums, since these are matching well defined enums in Wayland.

OTOH, we have 3 enums that define exactly the same values.

IMO Anchor and Gravity should be differents since they define different concept (the anchor is the point where we want to place the popup, while Gravity is how to position the popup at that point.)

Layer::Anchor and Popup::Anchor define the same concept, so idk if I should reuse the same type, or if I should keep them as-is.

Ottatop: I'd prefer keeping them separate. And actually, our layer anchor is being represented by an enum when it should really be a bitflag, I don't know why I implemented that way.

Ph4ntomas: > I'd prefer keeping them separate.

All right, then I'm keeping this as-is, I'll remove the comment.

> And actually, our layer anchor is being represented by an enum when it should really be a bitflag.

True, but the protocol doesn't really say what to do when setting opposite anchor points (e.g. Top|Bottom). At least the enum restricts choices to well known defined anchor points.

snowcap/api/protobuf/snowcap/widget/v1/widget.proto
```diff
@@ -216,6 +248,7 @@ message Container {
   optional bool clip = 8;
   WidgetDef child = 9;
   optional Style style = 10;
+  optional string id = 11;
```
Ph4ntomas, line 262: It occurs to me that adding an ID on containers might not be the best way to achieve this.

The `id` is needed to have *something* we can compute a anchor rect from, but idk if we want to use the Container widget and add an id on it.

This could break some config, if they build the container using the `Container {}` syntax without adding `..`.

However, assuming we want to create a new type for this, there are a few way to manage that:
- Exact duplicate
- Strip everything but the Id (meaning user may have to nest another container if they want the same theming property).
- Strip some of the container config (e.g. keep width and height ?).

IMO, we should either go with an empty `Bounded` type, or just add the id on the container.

snowcap/api/protobuf/snowcap/input/v1/input.proto
```diff
@@ -10,13 +10,18 @@ message Modifiers {
 }
 
 message KeyboardKeyRequest {
-  uint32 id = 1;
+  oneof target {
+    uint32 layer_id = 1;
+    uint32 popup_id = 2;
+  }
 }
```
Ph4ntomas, line 17: About that change:

This should not be API breaking:
- This is an internal message, only called from Layer at the moment, so only a client rebuild is needed here.
- The change itself is binary wire-safe, so there should not be any issue if an older client tries to send this message without rebuilding.

Let me know if I should change it regardless :)

Ph4ntomas:
> - [ ] Only allow one child popup ?

I'm not sure how to handle this one. At the moment, only one popup can get the keyboard grab, but since you can opt-out of keyboard grab, it could be used to have a few widget that don't have keyboard grab, but can still be displayed (and get pointer inputs).

OTOH, there are currently no mechanism to give keyboard grab to an open popup, and grab can only be taken if the parent has it.

For now, I'll add a way to replace the currently opened popup that you can opt-in, but I'm still undecided whether opening a new popup should be a hard error if there is already a popup with the same parent ID.

Ph4ntomas:
Okay implementation is done. 

I have a few commits I want to add to help with re-basing on main (like refactoring widget callback so the code isn't duplicated everywhere, which would decouple this PR from text-area & mouse-area).

snowcap/api/protobuf/snowcap/input/v1/input.proto
Ottatop: Should be fine

snowcap/api/protobuf/snowcap/widget/v1/widget.proto
Ottatop: I think it's fine to have it on the container. We're probably going to want to add ids to all the other iced widgets that allow you to set ids for operations, and we'd probably want to allow anchoring to those widgets as well.

Ph4ntomas: > and we'd probably want to allow anchoring to those widgets as well.

Agreed. 

I went with the minimum changes, and container are kinda transparent to add on an existing layout, but having that on other widget is a plus. I'll add that

snowcap/src/handlers.rs
```diff
@@ -289,30 +348,105 @@ impl Dispatch<WpFractionalScaleV1, WlSurface> for State {
...
+
+delegate_xdg_shell!(State);
+
+impl PopupHandler for State {
+    fn configure(
+        &mut self,
+        _conn: &Connection,
+        _qh: &QueueHandle<Self>,
+        popup: &smithay_client_toolkit::shell::xdg::popup::Popup,
+        _config: smithay_client_toolkit::shell::xdg::popup::PopupConfigure,
```
Ottatop:
```suggestion
        config: smithay_client_toolkit::shell::xdg::popup::PopupConfigure,
```

snowcap/src/lib.rs
```diff
@@ -103,6 +108,13 @@ pub fn start(stop_signal_sender: Option<tokio::sync::oneshot::Sender<SnowcapHand
                 deco.update(&mut state.runtime, state.tiny_skia.as_mut().unwrap());
                 deco.draw_if_scheduled();
             }
+
+            for popup in state.popups.iter_mut() {
+                if popup.initial_configure_received {
```
Ottatop: Is this check needed?

snowcap/src/state.rs
```diff
@@ -107,6 +112,9 @@ impl State {
         let foreign_toplevel_list: ExtForeignToplevelListV1 =
             globals.bind(&queue_handle, 1..=1, ()).unwrap();
 
+        // TODO: Decide whether we keep the full xdg_shell or just XdgWmBase
```
Ottatop: Probably fine to use xdg_shell for ease of implementation

snowcap/src/popup.rs
```diff
@@ -0,0 +1,562 @@
...
+
+                if anchor.is_within(&bounds.into()) {
+                    Some(iced::Rectangle {
+                        x: x as i32,
+                        y: y as i32,
+                        width: width as i32,
+                        height: height as i32,
+                    })
+                } else {
+                    None
```
Ottatop: TODO maybe clamp idk

snowcap/api/lua/snowcap/popup.lua
```diff
@@ -0,0 +1,387 @@
...
+
+---Build a handle to a popup's parent surface.
+---@enum snowcap.popup.parent
+local parent = {
+    ---Build a ParentHandle from a LayerHandle.
+    ---
+    ---@param handle snowcap.layer.LayerHandle
+    ---@return snowcap.popup.ParentHandle
+    Layer = function(handle)
+        return setmetatable({ layer = handle.id }, ParentHandle)
```
Ottatop, line 40: What's the purpose of this setmetatable?

snowcap/api/lua/snowcap/popup.lua
```diff
@@ -0,0 +1,387 @@
...
+    )
+
+    if err then
+        log.error(err)
+    end
+end
+
+---Do something on key press.
+---@param on_press fun(mods: snowcap.input.Modifiers, key: snowcap.Key)
+function PopupHandle:on_key_pressed(on_press)
```
Ottatop:
```suggestion
function PopupHandle:on_key_press(on_press)
```

snowcap/api/protobuf/snowcap/popup/v1/popup.proto
```diff
@@ -0,0 +1,118 @@
+syntax = "proto3";
+
+package snowcap.popup.v1;
+
+import "google/protobuf/empty.proto";
+import "snowcap/widget/v1/widget.proto";
+import "snowcap/operation/v1/operation.proto";
+
+// Reuse the one from layout ?
```
Ottatop:
```suggestion

```

snowcap/src/state.rs
```diff
@@ -155,19 +169,26 @@ impl State {
         loop_handle
             .insert_source(recv, move |event, _, state| match event {
                 calloop::channel::Event::Msg((id, msg)) => {
-                    let Some(layer) = state
+                    //TODO handle popup
```
Ottatop:
```suggestion

```

snowcap/src/handlers.rs
```diff
@@ -51,8 +56,13 @@ impl SeatHandler for State {
...
-    fn new_seat(&mut self, _conn: &Connection, _qh: &QueueHandle<Self>, _seat: WlSeat) {
-        // TODO:
+    fn new_seat(&mut self, _conn: &Connection, _qh: &QueueHandle<Self>, seat: WlSeat) {
+        // TODO: For now we only support one seat. This is good enough as most compositor only
+        // support one seat as well, but could be improved either by picking the best seat (the one
+        // with the most desirable capabilities), or having the user pick a seat by name.
+        if self.seat.is_none() {
+            self.seat = Some(seat);
+        }
     }
```
Ottatop, line 66: We also need to check if there's an existing seat right after we bind `SeatState`, this only gets called for new globals.

Ph4ntomas: > if there's an existing seat right after we bind SeatState

Since the loop hasn't have the chance to run when we bind the state, I wouldn't expect that to be possible, but I'll do the change (if only for the sake of correctness).

FWIW I've never had the case where the seat isn't added through this method.

snowcap/api/lua/snowcap/popup.lua
Ph4ntomas: Whoops

snowcap/api/lua/snowcap/popup.lua
Ph4ntomas: IIRC at some point I added a __tostring method to `ParentHandle` for debugging purposes. I can't remember if I decided it was not needed, or if it got scrapped by mistake.

Any preference ?

snowcap/src/handlers.rs
Ottatop: Initial seats are in the `GlobalList` on creation, we don't need a roundtrip. For whatever reason it *only* works for me if we do that.

snowcap/api/lua/snowcap/popup.lua
Ottatop: Guess we can leave the tostring in

snowcap/src/handlers.rs
Ph4ntomas: Ok, I've changed it, but I'm starting to wonder if it's actually needed at all. `SeatState` already holds the list of seats, so that variable could be replaced by `state.seat_state.seats.next()` when it's used.

I'm leaving it in for now.

snowcap/src/lib.rs
Ph4ntomas: Doesn't seem so, I've removed it.

snowcap/api/protobuf/snowcap/widget/v1/widget.proto
Ph4ntomas: I've addressed everything else. I'll add that asap, but if you'd rather have that in another PR, just let me know. I don't want to block that PR if you feel it's good to go.

Ottatop: Yea please open another PR for this

PR #406, opened 2026-01-14
backend: implement content-type based cursor throttling

anriha:
This PR implements the `wp_content_type_v1` protocol and refines the logic for cursor throttling on fullscreen windows with VRR enabled.

Currently, the compositor throttles cursor updates to ~24Hz for any fullscreen window to prevent VRR brightness flicker. While this is necessary for games and video playback, it creates a terrible user experience for native productivity applications (e.g., terminal emulators). On a 144Hz screen, a fullscreen terminal feeling locked to 24fps is jarring and unresponsive.

**Changes**
1. **Protocol Support**: Implemented `wp_content_type_v1` to allow clients to specify their content
2. Smart Throttling Logic: The render loop now uses the content type to decide whether to skip cursor-only updates:
    - Native Apps (Type: None): Unthrottled. This restores smooth 144Hz mouse movement for terminals, browsers, and editors.
    - Video / Photo: Throttled. Preserves correct frame pacing and power saving.
    - Game: Throttled. Ensures the refresh rate matches the game's rendering rate to prevent VRR gamma flicker/judder. If we decide to implement tearing support that should probably override this. 
    - XWayland: Throttled. Since XWayland windows do not currently support the content-type protocol, we assume fullscreen X11 windows are likely games and default to the "safe" legacy behavior to prevent flicker.

### DIFF
--- a/src/backend/udev.rs
+++ b/src/backend/udev.rs
@@ -1558,15 +1558,16 @@ impl Udev {
                 })
                 .unwrap_or(ContentType::None);
 
-            let should_throttle = match content_type {
-                ContentType::Photo | ContentType::Video | ContentType::Game => true,
+            // Only skip cursor updates for games.
+            //
+            // We don't do this for `Photo`s because they're static and don't stutter when
+            // the refresh rate jumps, nor for `Video`s because we don't expect the mouse
+            // to be moved much while video is playing.
+            //
+            // We assume xwayland windows may be a game for this purpose.
+            let should_skip = content_type == ContentType::Game || is_xwayland;
 
-                // If XWayland it is likely a game so we use legacy behavior
-                ContentType::None => is_xwayland,
-                _ => false,
-            };
-
-            if !too_long_since_last_present && cursor_over_fs_window && should_throttle {
+            if !too_long_since_last_present && cursor_over_fs_window && should_skip {
                 // FIXME: With a non-1 scale, the cursor no longer resides on the cursor plane,
                 // making this useless
 


src/backend/udev.rs
```diff
@@ -1537,9 +1545,28 @@ impl Udev {
...
+
+            let content_type = window_under
+                .and_then(|(win, _)| {
+                    win.toplevel()
+                        .map(|toplevel| get_surface_content_type(toplevel.wl_surface()))
+                })
+                .unwrap_or(ContentType::None);
+
+            let should_throttle = match content_type {
+                ContentType::Photo | ContentType::Video | ContentType::Game => true,
```
Ottatop:
```suggestion
                ContentType::Game => true,
```

I'd argue this should only be `Game`. If the content is `Photo` then there's no reason to skip cursor updates because photos are static and jacking up the refresh rate doesn't do anything, and if it's `Video` then you're usually not moving the mouse while you're watching so having the cursor be fluid is a better experience.

src/backend/udev.rs
```diff
@@ -1537,9 +1545,28 @@ impl Udev {
...
+            let is_xwayland = window_under.is_some_and(|(win, _)| win.is_x11());
+
+            let content_type = window_under
+                .and_then(|(win, _)| {
+                    win.toplevel()
+                        .map(|toplevel| get_surface_content_type(toplevel.wl_surface()))
+                })
+                .unwrap_or(ContentType::None);
+
+            let should_throttle = match content_type {
```
Ottatop: I'd rename this to `should_skip`

src/backend/udev.rs
```diff
@@ -1537,9 +1545,28 @@ impl Udev {
...
+                ContentType::Photo | ContentType::Video | ContentType::Game => true,
+
+                // If XWayland it is likely a game so we use legacy behavior
+                ContentType::None => is_xwayland,
+                _ => false,
+            };
```
Ottatop:
```suggestion
            // Only skip cursor updates for games.
            //
            // We don't do this for `Photo`s because they're static and don't stutter when
            // the refresh rate jumps, nor for `Video`s because we don't expect the mouse
            // to be moved much while video is playing.
            //
            // We assume xwayland windows may be a game for this purpose.
            let should_skip = content_type == ContentType::Game || is_xwayland;
```

I'd argue we should only do this for games.

src/backend/udev.rs
```diff
@@ -1537,9 +1545,28 @@ impl Udev {
...
+
+            if !too_long_since_last_present && cursor_over_fs_window && should_throttle {
```
Ottatop:
```suggestion
            if !too_long_since_last_present && cursor_over_fs_window && should_skip {
```

anriha:
I assumed that the vrr flicker was the main concern, but if it isn't this makes more sense.

PR #407, opened 2026-01-16
fix(nix): UWSM session naming and introspection

cassandracomar:
this PR addresses a minor annoyance. by using a script in the nix store to pass arguments to pinnacle, we've made the session launched by UWSM possess an incredibly obtuse name. this makes it difficult to e.g. check session status with the UWSM binary or to check session logs in journald. I also just missed that the extra script was unnecessary because there was an `extraArgs` option we could just set instead.

### DIFF
--- a/nix/modules/nixos/default.nix
+++ b/nix/modules/nixos/default.nix
@@ -1,55 +1,53 @@
-{pkgs, config, lib, ...}:
-let
+{
+  pkgs,
+  config,
+  lib,
+  ...
+}: let
   cfg = config.programs.pinnacle;
-in with lib.options; {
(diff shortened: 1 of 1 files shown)


Ottatop:
> are we comfortable with Pinnacle's stability at this point or do we want to let it cook a bit longer?

I think we're good on stability. Before you upstream, I fired up a NixOS vm a few days ago and it seems you need to run `cargo build` to generate a Cargo.lock in the config directory before `buildRustConfig` can work, is that intended? I suppose it's either that or `ln` a new, possibly changing lockfile every rebuild which doesn't seem very reproducible.

PR #408, opened 2026-01-17
Snowcap: Add background for widget

Ph4ntomas:
Following the discussion in #369, here is the PR to add support for Background in widgets.

TODO:
- [x] Add background message and ser/deser support
- [x] Deprecate protobuf field and replace them with the background
- [x] Add conversion from background_color to background while the field is deprecated
- [x] Deprecate usage in rust api
- [x] Deprecate usage in lua api

### DIFF
--- a/snowcap/src/api/widget/v1.rs
+++ b/snowcap/src/api/widget/v1.rs
@@ -319,12 +319,12 @@ pub fn widget_def_to_fn(def: WidgetDef) -> Option<ViewFn> {
                             s.vertical_rail.scroller.border = FromApi::from_api(scroller_border);
                         }
                         if let Some(background) = background {
-                            s.horizontal_rail.background = FromApi::from_api(background);
+                            s.vertical_rail.background = TryFromApi::try_from_api(background).inspect_err(|e| tracing::error!("{e}")).ok();
                         }
                         if let Some(scroller_background) = scroller_background
-                            && let Some(background) = FromApi::from_api(scroller_background)
+                            && let Some(background) = TryFromApi::try_from_api(scroller_background).inspect_err(|e| tracing::error!("{e}")).ok()
                         {
-                            s.horizontal_rail.scroller.background = background;
+                            s.vertical_rail.scroller.background = background;
                         }
                     }
                     if let Some(h_rail) = style.as_ref().and_then(|s| s.horizontal_rail.clone()) {
@@ -355,10 +355,10 @@ pub fn widget_def_to_fn(def: WidgetDef) -> Option<ViewFn> {
                             s.horizontal_rail.scroller.border = FromApi::from_api(scroller_border);
                         }
                         if let Some(background) = background {
-                            s.horizontal_rail.background = FromApi::from_api(background);
+                            s.horizontal_rail.background = TryFromApi::try_from_api(background).inspect_err(|e| tracing::error!("{e}")).ok();
                         }
                         if let Some(scroller_background) = scroller_background
-                            && let Some(background) = FromApi::from_api(scroller_background)
+                            && let Some(background) = TryFromApi::try_from_api(scroller_background).inspect_err(|e| tracing::error!("{e}")).ok()
                         {
                             s.horizontal_rail.scroller.background = background;
                         }
@@ -467,7 +467,9 @@ pub fn widget_def_to_fn(def: WidgetDef) -> Option<ViewFn> {
                     }
 
                     if let Some(background) = background_clone.clone() {
-                        style.background = FromApi::from_api(background);
+                        style.background = TryFromApi::try_from_api(background)
+                            .inspect_err(|e| tracing::error!("{e}"))
+                            .ok();
                     }
 
                     style
@@ -554,7 +556,9 @@ pub fn widget_def_to_fn(def: WidgetDef) -> Option<ViewFn> {
                             }
 
                             if let Some(background) = style.background {
-                                s.background = FromApi::from_api(background);
+                                s.background = TryFromApi::try_from_api(background)
+                                    .inspect_err(|e| tracing::error!("{e}"))
+                                    .ok();
                             }
                         }
 
@@ -886,7 +890,9 @@ impl FromApi<widget::v1::container::Style> for iced::widget::container::Style {
         }
 
         if let Some(background) = background {
-            ret.background = FromApi::from_api(background);
+            ret.background = TryFromApi::try_from_api(background)
+                .inspect_err(|e| tracing::error!("{e}"))
+                .ok();
         }
 
         ret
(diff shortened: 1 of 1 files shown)


snowcap/api/lua/snowcap/widget.lua
```diff
@@ -200,6 +200,65 @@ local alignment = {
...
+    ---@type fun(radians: number, stops: snowcap.widget.gradient.ColorStop[]): snowcap.widget.Background
+    Linear = function(radians, stops)
+        ---@type snowcap.widget.gradient.Linear
+        local linear = { radians = radians, stops = stops or {} }
+
+        ---@type snowcap.widget.Gradient
+        local gradient = { linear = linear }
+
+        return { gradient = gradient }
+    end,
```
Ottatop: Should we have both of these? If iced adds more gradient types and we add a `Radial` or `Conical` then `Gradient` seems redundant. I think I'd rather just have `Gradient` only for now, even if it means more nesting on construction.

snowcap/src/api/widget/v1.rs
```diff
@@ -845,3 +916,53 @@ impl FromApi<widget::v1::Border> for iced::Border {
         ret
     }
 }
+
+impl FromApi<widget::v1::Background> for Option<iced::Background> {
```
Ottatop: Nit: this should be `TryFromApi` and the call sites call .ok(), but I also understand that that loses the tracing error unless you also inline error handling. Maybe make a wrapper function that has the same functionality, something about a `From -> Option` just for error logging doesn't feel right

snowcap/api/rust/src/widget.rs
```diff
@@ -371,6 +372,162 @@ impl From<Radius> for widget::v1::Radius {
...
+    fn from(gradient: Gradient) -> Self {
+        Self::Gradient(gradient)
+    }
+}
+
+impl From<Linear> for Background {
+    fn from(linear: Linear) -> Self {
+        Self::Gradient(Gradient::Linear(linear))
+    }
+}
```
Ottatop, line 400: Guess I'm ok with this impl in the context of the above comment but if you want to remove it to keep "parity" with the Lua side I don't mind

snowcap/api/rust/src/widget.rs
```diff
@@ -371,6 +372,162 @@ impl From<Radius> for widget::v1::Radius {
...
+    }
+
+    /// Adds a new [`ColorStop`], defined by an offset and a color.
+    ///
+    /// `offset`s not within the 0.0..=1.0 range will be ignored.
+    /// Any stop added after the 8th will be ignored.
+    ///
+    /// If a new stop with an equivalent offset is added, it will replace the previous one.
+    /// Equivalence is defined by `||old - new|| <= f32::EPSILON`
+    pub fn add_stop(self, offset: f32, color: Color) -> Self {
```
Ottatop, line 459: We should just push a `ColorStop` and let iced handle this logic instead of duplicating it client-side. Not sure it's worth having the `warn`s for debugging.

snowcap/src/api/widget/v1.rs
Ph4ntomas: The reasoning here was that most (if not every) time we call this, we want an Option anyway (because of iced API), so I removed the `Try`, but I agree it's unusual.

> Maybe make a wrapper function that has the same functionality

Yeah I guess that would work, will do.

snowcap/api/lua/snowcap/widget.lua
Ph4ntomas: IMO `Gradient` is redundant, so if one should go, it should be this one since the `gradient` field is an implementation details

One of the reason for Linear to be here is that it mirror iced's (and rust api) `From<Linear> for Background` trait implementation, which feels like the "intended" way to build a Nackground, while the `Gradient` enum is just there to hold the variants.

Ottatop: In that case let's remove `Gradient` then.

snowcap/api/rust/src/widget.rs
Ph4ntomas: > Not sure it's worth having the warns for debugging.

Well `tracing` is currently broken by default on client side, because the format it produces is not the one handled by pinnacle, but I have it working on my config by replacing the default logger/formatter (I was planing to open a PR for that at some point)

> We should just push a ColorStop and let iced handle this logic instead of duplicating it client-side.

Fair enough, I'm pretty sure this was some experiment I forgot to remove anyway. I'll note that iced impl have several issue tho:
- the log will be coming from the server
- Pushing stops out of order doesn't quite have the correct behavior iirc (iced replaces at the index instead of inserting), but that should be fixed upstream, not worked around client side
- Iced silently ignore the case where index > 8.

I'd rather keep the logging if it's ok for you, because I don't like silently discarding stuff, and I'd rather have the warning coming from the config than one of the server's library.

snowcap/api/rust/src/widget.rs
Ph4ntomas: I kinda want to keep it, as I don't see a point in making the code 'harder' to write.

Ph4ntomas: Since we're keeping Linear in rust, I'm resolving this

snowcap/api/rust/src/widget.rs
Ph4ntomas: After ~5 additional minutes, I went ahead and remove the logging as well. Lua doesn't do anything special, let's keep it simple.

snowcap/src/api/widget/v1.rs
Ph4ntomas: Did not go with a wrapper function. it's just a `.map_err().ok()`, so it's still readable imo.

snowcap/api/lua/snowcap/widget.lua
```diff
@@ -200,6 +200,61 @@ local alignment = {
...
+---@field color snowcap.widget.Color?
+---Interpolate between several colors.
+---@field gradient snowcap.widget.Gradient?
+
+---Builders for `Background`.
+---@class snowcap.widget.background
+---Builds a `Background` from a solid `Color`.
+---@field Color fun(color: snowcap.widget.Color): snowcap.widget.Background
+---Builds a `Background` from a generic `Gradient`.
+---@field Gradient fun(gradient: snowcap.widget.Gradient): snowcap.widget.Background
```
Ottatop:
```suggestion
```

snowcap/src/api/widget/v1.rs
```diff
@@ -282,56 +283,85 @@ pub fn widget_def_to_fn(def: WidgetDef) -> Option<ViewFn> {
...
                         if let Some(scroller_color) = scroller_color {
+                            tracing::warn!("Rail::scroller_color is deprecated. Use Rail::scroller_background instead.");
                             s.vertical_rail.scroller.background =
                                 iced::Background::Color(FromApi::from_api(scroller_color));
                         }
                         if let Some(scroller_border) = scroller_border {
                             s.vertical_rail.scroller.border = FromApi::from_api(scroller_border);
                         }
+                        if let Some(background) = background {
+                            s.vertical_rail.background = TryFromApi::try_from_api(background).map_err(|e| tracing::error!("{e}")).ok();
```
Ottatop: `inspect_err` is more appropriate but also it literally doesn't matter

Ottatop (changes requested):

snowcap/api/lua/snowcap/widget.lua
Ph4ntomas: Thanks !

snowcap/src/api/widget/v1.rs
Ph4ntomas: Let's go with that, I just wasn't aware that function existed.

Ottatop (approved):
Thank you!

PR #413, opened 2026-01-24
specify drm devices via env var

rustysec:
Problem: when running pinnacle, the dGPU is constantly kept awake leading to high power usage and faster battery drain.

This PR allows the user to instruct pinnacle and snowcap of GPU preference. Similar to how wlroots allows users to specify with `WLR_DRM_DEVICES`, I added some parsing around `PINNACLE_DRM_DEVICES`. If the `/dev/dri` device is not in this list, it will not be added to pinnacle's renderers.

Example:

```bash
PINNACLE_DRM_DEVICES="/dev/dri/card1:/dev/dri/renderD128" pinnacle-session
```

For snowcap, the default wgpu selector is `None` which allows it to pick whatever it wants. I added the built in `.from_env()` call to access the `WGPU_POWER_PREF` variable.

example:

```bash
WGPU_POWER_PREF=low pinnacle-session
```

my current launch script looks like this:

```bash
PINNACLE_DRM_DEVICES="/dev/dri/card1:/dev/dri/renderD128" WGPU_POWER_PREF=low  pinnacle-session
```
this allows my dGPU to suspend properly when not in use by some background application.

### DIFF
--- a/src/backend/udev.rs
+++ b/src/backend/udev.rs
@@ -154,17 +154,26 @@ impl Udev {
         let (session, notifier) = LibSeatSession::new()?;
 
         // Get the primary gpu
-        let primary_gpu = if let Some(device) = std::env::var("PINNACLE_DRM_DEVICES")
+        let primary_gpu = std::env::var("PINNACLE_DRM_DEVICES")
             .ok()
             .and_then(|var| var.split(":").next().map(String::from))
-        {
-            info!("attempting to set primary device as {device}");
+            .and_then(|device| {
+                info!("Attempting to set primary device as {device}");
 
-            DrmNode::from_path(device)?
-                .node_with_type(NodeType::Render)
-                .expect("unable to set primary drm device")?
-        } else {
-            udev::primary_gpu(session.seat())
+                match DrmNode::from_path(device) {
+                    Ok(drm_node) => drm_node
+                        .node_with_type(NodeType::Render)
+                        .and_then(|node| node.ok()),
+                    Err(err) => {
+                        warn!("Unable to use specified device {err}, falling back");
+                        None
+                    }
+                }
+            });
+
+        let primary_gpu = match primary_gpu {
+            Some(primary) => primary,
+            None => udev::primary_gpu(session.seat())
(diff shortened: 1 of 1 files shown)


src/backend/udev.rs
```diff
@@ -2115,3 +2134,18 @@ fn handle_pending_screencopy<'a>(
         }
     }
 }
+
+#[allow(dead_code, unused_variables)]
```
Ottatop:
```suggestion
```

src/backend/udev.rs
```diff
@@ -2115,3 +2134,18 @@ fn handle_pending_screencopy<'a>(
...
+fn should_use_drm_device<P: AsRef<Path>>(device_path: P) -> bool {
+    if let Ok(var) = std::env::var("PINNACLE_DRM_DEVICES") {
+        let device_path = device_path.as_ref();
+
+        return var
+            .split(":")
+            .map(String::from)
+            .map(std::path::PathBuf::from)
```
Ottatop:
```suggestion
            .map(Path::new)
```

src/backend/udev.rs
rustysec: oof, left over from debugging.

src/backend/udev.rs
```diff
@@ -154,22 +154,32 @@ impl Udev {
...
-        info!("Using {} as primary gpu", primary_gpu);
+        let primary_gpu = if let Some(device) = std::env::var("PINNACLE_DRM_DEVICES")
+            .ok()
+            .and_then(|var| var.split(":").next().map(String::from))
+        {
+            info!("attempting to set primary device as {device}");
+
+            DrmNode::from_path(device)?
+                .node_with_type(NodeType::Render)
+                .expect("unable to set primary drm device")?
```
Ottatop: Just realized that this will prevent the compositor from starting if it can't find the first listed device. This should fall back to the stuff below and emit a `warn!` on failure instead of returning out of the whole function

rustysec: added some additional handling and `warn!` in case the device isn't there. if it isn't, it will just fallback to the original code.

PR #414, opened 2026-01-24
snowcap: Make programs composable

Ottatop:
This PR changes `Program` to enable composition within other `Program`s.

This is pretty much just folding [Glacier](https://git.sr.ht/~phantomas/glacier)'s `Widget` and `TryWithEmitter` traits into the `Program` trait (with Lua doing similar), because there's no way to easily embed something stateful into a `Program` without just recreating Glacier. It also introduces signals.

There are some differences.

- I renamed `Emitter` to `Signaler`
- Instead of a toplevel surface handling signals, the `new_widget` functions connect to `RedrawNeeded` and `Message` signals
- The Rust API introduces a `UniversalMsg` as a general catch-all message
    - Messages that impl `Universal` also impl `From<T>` and `Into<Option<T>>` to unify messages within a program tree

@Ph4ntomas I stole a bunch of stuff from Glacier so I'm going to need your permission to re-license to MPL. Also any input on this PR is helpful.

### DIFF
--- a/snowcap/api/rust/src/widget/signal.rs
+++ b/snowcap/api/rust/src/widget/signal.rs
@@ -29,3 +29,9 @@ impl<Msg> From<Msg> for Message<Msg> {
 }
 
 impl<Msg: Clone + 'static> Signal for Message<Msg> {}
+
+/// Notifies that a widget closed.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct Closed;
+
+impl Signal for Closed {}
--- a/snowcap/api/lua/snowcap/widget/signal.lua
+++ b/snowcap/api/lua/snowcap/widget/signal.lua
@@ -6,6 +6,10 @@
 ---
 ---@enum snowcap.widget.signal
 return {
+    ---Notifies that a redraw is needed.
     redraw_needed = "widget::redraw_needed",
+    ---Emits a message that will update widgets.
     send_message = "widget::send_message",
+    ---Notifies that a widget closed.
+    closed = "widget::closed",
 }


snowcap/api/rust/src/widget.rs
```diff
@@ -609,12 +615,77 @@ impl From<LineHeight> for widget::v1::LineHeight {
...
+/// A `Program` builds a widget for display by Snowcap and updates itself
+/// from messages generated from interactions with the widget and from other
+/// external sources.
 pub trait Program {
+    /// The type of messages that this widget program receives.
     type Message;
 
+    /// Updates this widget program with the received message.
+    ///
+    /// Returns whether the view should be recreated.
```
Ottatop: Ignore this doc comment, was trying to enable updates that don't trigger a redraw without big changes and left it in

snowcap/api/rust/src/signal.rs
```diff
@@ -0,0 +1,170 @@
...
+///
+/// [`Signaler`]s holds handlers for signals in a type-erased way. Other types can
+/// [connect] and [disconnect] handlers, or [emit] signals.
+///
+/// [connect]: Signaler::connect
+/// [disconnect]: Signaler::disconnect
+/// [emit]: Signaler::emit
+#[derive(Default, Clone, Debug)]
+pub struct Signaler {
+    entries: Arc<Mutex<HashMap<TypeId, Box<dyn Any + Sync + Send>>>>,
```
Ottatop: This is now `TypeId`, but now there isn't great debug information for the types. Unsure if allowing all types is good or if it should be opt-in with a `Signal` trait like it is in Glacier

snowcap/api/lua/snowcap/layer.lua
```diff
@@ -116,7 +117,26 @@ function layer.new_widget(args)
...
+            layer_id = layer_id,
+        })
+
+        if err then
+            log.error(err)
+        end
+    end
+
+    args.program:connect(widget_signal.redraw_needed, update_on_msg)
+    args.program:connect(widget_signal.send_message, update_on_msg)
```
Ottatop, line 145: Realizing these don't return a handler policy, same with the deco stuff

snowcap/api/rust/src/signal.rs
Ph4ntomas: On glacier, I've used the trait because I wanted to restrict somewhat what type can be passed as a signal to force a more 'conscious' use. The reasoning is that while it's useful in this context, type-erasure sidestep Rust strict typing, so having that opt-in is IMO a better design because it's made more explicit.

The signal_name as key came a bit later, but `TypeId` might be better in this context. The trait could still be used for debug information tho

snowcap/api/lua/snowcap/layer.lua
Ph4ntomas: This is less of an issue on lua. 

From what I can see you've used my code, which makes discarding opt-in, so returning nil (or not returning anything) is equivalent to 'Keep'. I'm pretty sure I've done the same in a few of Glacier's functions. 
(But tbh it make sense for the built-in signal handler to be more correct too.)

snowcap/api/rust/src/signal.rs
Ottatop: > On glacier, I've used the trait because I wanted to restrict somewhat what type can be passed as a signal to force a more 'conscious' use. The reasoning is that while it's useful in this context, type-erasure sidestep Rust strict typing, so having that opt-in is IMO a better design because it's made more explicit.

I agree, re-added the Signal trait. The key is now (TypeId, &'static str) to prevent any accidental collisions.

snowcap/api/rust/src/widget/base.rs
```diff
@@ -0,0 +1,47 @@
...
+
+use crate::signal::Signaler;
+
+/// A building block providing common widget functionality.
+///
+/// This enables widgets to uniquely identify themselves and
+/// connect and emit signals.
+#[derive(Debug)]
+pub struct WidgetBase {
+    widget_type: String,
```
Ottatop, line 11: I believe this string is only used for `Display`, not sure how useful that is

snowcap/api/rust/src/widget/signal.rs
```diff
@@ -0,0 +1,31 @@
+use crate::signal::Signal;
+
+/// Notifies that a redraw is needed.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct RedrawNeeded;
+
+impl Signal for RedrawNeeded {}
```
Ottatop, line 7: Might want to make a derive for this down the line

snowcap/api/rust/src/widget/base.rs
Ph4ntomas: > not sure how useful that is.

From my experience with Glacier, it's useful when debugging more complex Widgets (e.g. a tag list and its nested tags)

snowcap/api/rust/src/widget.rs
```diff
@@ -628,12 +635,164 @@ impl From<Wrapping> for widget::v1::Wrapping {
...
+pub trait Source {
+    /// The type of messages that this source receives.
+    type Message;
+
+    /// Registers this source with the provided widget's [`Signaler`].
+    fn register(&mut self, signaler: Signaler);
 
+    /// Updates this source with the received message.
+    fn update(&mut self, msg: Self::Message);
+}
```
Ph4ntomas: I'm not really sure I understand how `Source` are meant to work.

Is the `register` function meant to clone the Signaler for later use ? 
If so, I'm not a huge fan since that introduces a way to leak the `Signaler` state if misused, especially since there's no current way to know a `Widget` is being discarded or signals are being deleted.

The `update` function is also a bit weird IMO. I would expect a message `Source` to be stateless from the Program perspective (but that's more a case of "I'm not sure I understand the use-case").

From the comment on `Program::update`, the source are meant to be held by the `Program`, but there's nothing enforcing that, which increase the risk of leaking the `Signaler`.

I could kinda see the point for #415 to have a `Source` maintain the tasks list state, and have every widget connect to that source. The widgets would then receive the initial state upon registration, then will only get updates when the state changes (meaning you would have only one channel that speak to the server instead of one per widgets).

But that would mean reversing the logic, so:
```rust
/// A source of data that can be used in widget programs.
pub trait Source {
    /// The type of signal that this source emit
    type Signal;
    
    /// Maybe have a type for the initial state

    /// Returns the initial state as a vec of `Signal`, and the `Signaler` to get future updates
    fn register(&mut self) -> (&Signaler, Vec<Self::Signal>);
}
```
or
```rust
/// A source of data that can be used in widget programs.
pub trait Source {
    /// The type of signal that this source emit.
    type Signal;

    // Maybe add a type for the initial state

    /// Connect the program to this signaler signal.
    fn connect<P, F, R>(&self, program: &mut P, initial_state_processor: F, update_processor: R)
        where
            P -> Program,
            F -> FnOnce(&mut P, Vec<Signal>)
            R -> Fn(Signal)
    ;
    
    // Maybe add a function for stateless connection without the initial_state_processor, or split the trait into stateless and stateful Sources
}
```

Ph4ntomas: OTOH, if the source are meant to be fully contained inside Program, I don't think there's a need for a trait. The `Program` will already need to store the object implementing `Source`, so there's no reason to limit ourself to the `Source` since the full type is known.

It would be useful for the task list to be able to make use of a generic `Source<Signal: TaskChanges>` that could either get its data from wlr-foreign-toplevel-management or xdg-foreign-toplevel-list, but I don't really see how the current impl would make it easier than having a `TaskSource` trait with a more specialized API

snowcap/api/rust/src/widget/signal.rs
Ph4ntomas: I could adapt the one from Glacier

snowcap/api/rust/src/widget.rs
Ottatop: > Is the `register` function meant to clone the Signaler for later use ?

Clone or otherwise move, yes. As an example:
```rust
struct WindowTitle {
    pub name: String,
}

impl Source for WindowTitle {
    type Message = String;

    fn register(&mut self, signaler: Signaler) {
        pinnacle_api::window::connect_signal(WindowSignal::TitleChanged(Box::new(move |_win, title| {
            signaler.emit(Message(title.to_string()));
        })));
    }

    fn update(&mut self, msg: String) {
        self.name = msg;
    }
}
```

Then some other `Program` creates an instance of `WindowName`, calls `source.register()` with its signaler, and stores it. Then when a window's title changes and a message gets emitted, the program calls `source.update()` with the message. When `update` on the program is called it can use the title from the source.

> If so, I'm not a huge fan since that introduces a way to leak the `Signaler` state if misused, especially since there's no current way to know a `Widget` is being discarded or signals are being deleted.

That's... *definitely* a problem lol. Perhaps it would be a good idea to make all `Signaler`s apart from the one in the `WidgetBase` weak.

> The `update` function is also a bit weird IMO. I would expect a message `Source` to be stateless from the Program perspective (but that's more a case of "I'm not sure I understand the use-case").

I guess "source" isn't a great name for it. It's supposed to be stateful (as seen above).

> From the comment on `Program::update`, the source are meant to be held by the `Program`, but there's nothing enforcing that, which increase the risk of leaking the `Signaler`.

I'm not sure how we *would* enforce that tbh. Leaking the signaler is much less of an issue if we make it weak as per above.

> OTOH, if the source are meant to be fully contained inside Program, I don't think there's a need for a trait.
>
> It would be useful for the task list to be able to make use of a generic Source<Signal: TaskChanges> that could either get its data from wlr-foreign-toplevel-management or xdg-foreign-toplevel-list, but I don't really see how the current impl would make it easier than having a TaskSource trait with a more specialized API

Yea I suppose ditching this `Source` trait for more specialized traits is desirable seeing that this one doesn't really provide any extra functionality.

snowcap/api/rust/src/widget/base.rs
Ottatop: Guess we can leave it in then

snowcap/api/rust/src/widget/signal.rs
Ottatop: PR welcomed in that case

snowcap/api/rust/src/widget.rs
Ph4ntomas: > That's... definitely a problem lol. Perhaps it would be a good idea to make all Signalers apart from the one in the WidgetBase weak.

I don't think it's required (you'd need to promote it to a Signaler to use it anyway, so it's kinda a useless step), but the API should definitely not encourage passing it around or storing it. Maybe it would be better to just return it by reference. That way it's obvious from reading the API that you're not supposed to hold onto it, and the internals can clone it when/if needed to make the borrow-checker happy.

> Yea I suppose ditching this Source trait for more specialized traits is desirable seeing that this one doesn't really provide any extra functionality.

There might be a case to be made to reduce the amount of open stream with the server and reduces the amount of memory consumed using a stateful object to hold shared state and notify of said state changes. I'm not sure we currently need that, and that would be kinda the opposite to what this trait does anyway.

Ottatop: > I don't think it's required (you'd need to promote it to a Signaler to use it anyway, so it's kinda a useless step)

Yea I suppose so

> Maybe it would be better to just pass it by reference.

I mean it's going to need to call `emit` after the fact which will require storage anyway, passing a reference just introduces one more forced clone unless it stores the reference which would suck in terms of lifetimes.

For now let's leave everything as is, we can assume that the source is responsible for cleanup of uses of the signaler on drop or something.

> There might be a case to be made to reduce the amount of open stream with the server and reduces the amount of memory consumed using a stateful object to hold shared state and notify of said state changes. I'm not sure we currently need that, and that would be kinda the opposite to what this trait does anyway.

Yea this doesn't seem like a huge issue. I'll remove the trait, we can figure something out later.

snowcap/api/lua/snowcap/layer.lua
Ottatop: In that case guess I'll leave it as is, no reason to add `return policy.Keep`

snowcap/api/rust/src/widget/signal.rs
```diff
@@ -0,0 +1,35 @@
...
+    fn from(value: Msg) -> Self {
+        Self::new(value)
+    }
+}
+
+impl<Msg: Clone + 'static> Signal for Message<Msg> {}
+
+/// Notifies that a widget closed.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct Closed;
```
Ph4ntomas, line 35:
```suggestion
pub struct Closed;

impl Signal for Closed {}
```

Ottatop: oops

Issue #412, opened 2026-01-22
[Snowcap] Task List RFC


Ph4ntomas:
I've been tinkering with wlr-foreign-toplevel-management, and I'm finally at a stage where I have something to show.

My general idea was to implement a server side widget that would work a bit like a mouse-area. It's goal is not to present anything, but to allow for the state to be notified of changed in the task.

https://github.com/user-attachments/assets/177c51d5-938b-4c86-b8f5-ea5529c25e02

The widget currently exposes 3 signals sources: 
Enter: A new task is being tracked by the widget.
Leave: A task stopped being tracked by the widget.
Update: A tracked task state has changed.

The widget additionally take a single child which will hold the content to be rendered (this means the config choose how the actual task is being rendered.

### How it works:
1. Snowcap is notified of the state change for `wlr-foreign-toplevel-management`.
2. A set of custom [operation](https://docs.rs/iced_core/0.14.0/iced_core/widget/operation/index.html) are broadcasted to every surfaces. 
3. The widget react to this operation and prepare message that are stored in its internal state.
4. The pending changes are converted into messages next time the update function is called
5. After the broadcast is done, wlr-foreign-toplevel-management functions request a redraw.

### Planned features (assuming I/we proceed with this):
- Make the widget output aware, so it only notify for toplevels on its output (and make this a config option)
- Add a way to interact with the tasks being displayed. The idea here would be to implement additional operation exposed to the config to call some methods on wlr-foreign-toplevel-management handles. The preliminary work is already done, as the widget hold on weak handles, and transmit a unique id as part of the messages it emit.

### Notes & caveat:
I've base the POC on wlr-foreign-toplevel-management because it's supported by most compositor and can actually control the foreign toplevels. 
However this protocol has a few limitation:
  - No support for workspace, the best you can do is filter per output.
  - No unique identifier per toplevel (so no way to tie it back to pinnacle config).

`ext-foreign-toplevel-list` doesn't have a way to act on the toplevel [yet](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/196), but we get an identifier that we can the tie back to the config using `WindowHandle:foreign_toplevel_list_identifier()`

Long term, I think we should move to `foreign-toplevel-state` (assuming it gets merged), but for the time being, there are a few things that can be done:
- Have a `WlrTaskList` widget (essentially what I have a poc for), that only cares about the wlroot implementation, even if it have some limitation.
[...]

PR #415, opened 2026-01-24
Snowcap: Add WlrTaskList support

Ph4ntomas:
As mentioned in #412, here is a (still wip) widget to support WlrTaskList.

A few things are still missing:
- [ ] Rust support
- [ ] Per-output filtering (I'm still unsure if I want that to be done server-side or client-side)
- [x] Operations to call wlr-foreign-toplevel-management methods

### DIFF
--- a/snowcap/src/handlers.rs
+++ b/snowcap/src/handlers.rs
@@ -234,6 +234,8 @@ impl CompositorHandler for State {
         surface: &WlSurface,
         output: &wl_output::WlOutput,
     ) {
+        use crate::widget::output;
+
         let Some(layer) = self
             .layers
             .iter_mut()
@@ -243,6 +245,9 @@ impl CompositorHandler for State {
         };
 
         layer.wl_output = Some(output.clone());
+
+        let mut oper = output::operation::enter_output(output.clone());
+        layer.operate(&mut oper);
 
         let Some(output_info) = self.output_state.info(output) else {
             return;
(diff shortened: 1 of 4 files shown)


Ottatop:
> wlr-foreign-toplevel-management requires a seat to activate a toplevel.

> so we might as well keep things simple for the time being, and assume a single seat

Yea let's not overcomplicate things here. If we *are* going to add multi-seat it's going to be fairly far into the future.

---

>  [ ] Per-output filtering (I'm still unsure if I want that to be done server-side or client-side)

Think I'd want this done client-side in case anyone wants an all-encompassing tasklist.

---

> I'm also unsatisfied with the current implementation. While everything works (except per-output filtering), I'm currently capturing a Weak WlSeat on widget creation, the state for wlr-foreign-toplevel-management and a bunch of toplevel handles. I wonder if the protocol state should act as a 'service' holding knowledge of the Seat and keeping toplevel handles internally (only exposing an id to the widgets). This would simplify the widget implementation since it wouldn't need to hold on to both toplevel handles and wlseat.

Before I think about this, I'll ask: why is this implemented as an in-tree widget? I get that it allows us to plug in to the preexisting widget framework, but from what I can see the *only* purpose of this is to generate messages and respond to toplevel requests. I don't see why this couldn't be something like a server-side signal with the same `Added`, `Changed`, and `Closed` events along with `WindowHandle`-like methods. Plus if it's a widget there's the downside of having to either store toplevel handles within the widget or wiggle information from the widget to the server.

PR #417, opened 2026-01-27
Gestures

rustysec:
I implemented gesture handing and configuration, although I'm in definite need of some feedback on how I went about this. Currently the API assumes these are swipe gestures but I would be more than happy to handle pinch and hold if that makes sense.

Outstanding items:
* [x] use mod keys correctly
* [ ] ~~possibly refactor in `src/input/binds.rs` to match `Mousebinds` and `Keybinds` implementation more closely~~
* [x] add pinch and hold gesture handling
* [ ] add lua api

Some basic examples of how I'm using gestures from my config:

```rust
// Three fingers up launches fuzzel
input::gesturebind(Mod::empty(), GestureDirection::Up, GestureFingers::Three).on_finish(|| {
    Command::with_shell(["bash", "-c"], "fuzzel || pkill -9 fuzzel").spawn();
});

// Four fingers down spawns a terminal (just for testing, honestly)
input::gesturebind(Mod::empty(), GestureDirection::Down, GestureFingers::Four).on_finish(
    || {
        Command::with_shell(["bash", "-c"], "foot").spawn();
    },
);

// Three fingers left switches to the "next" tag
input::gesturebind(Mod::empty(), GestureDirection::Left, GestureFingers::Three).on_finish(
    || {
        if let Some(output) = pinnacle_api::output::get_focused()
            && let Some(current) = output.active_tags().next()
            && let Some(pos) = output.tags().position(|tag| tag == current)
            && let Some(next) = output.tags().nth(pos + 1)
        {
            next.switch_to();
        }
    },
);

// Three fingers right moves to the "previous" tag
input::gesturebind(Mod::empty(), GestureDirection::Right, GestureFingers::Three).on_finish(
    || {
        if let Some(output) = pinnacle_api::output::get_focused()
            && let Some(current) = output.active_tags().next()
            && let Some(pos) = output.tags().position(|tag| tag == current)
            && let Some(next) = output.tags().nth(pos.saturating_sub(1))
        {
            next.switch_to();
        }
    },
);
```


rustysec:
No, I can remove the restriction.

Ph4ntomas:
Regarding the Direction, is it possible to expand it to 8 directions ? We can't realistically have arbitrary deltas here (or rather not with the bind api), but imo it would be nice to support simple diagonals

Ph4ntomas:
Tbh I don't necessarily have a preference UDLR and NSWE (though maybe i'd personnally prefer UpLeft to UpAndLeft).

rustysec:
> Not necessarilly for this PR, but do you know if that's possible ? (It requires knowing the dimension of the touchpad at the very least)

I honestly haven't looked into it, but I certainly can.

api/protobuf/pinnacle/input/v1/input.proto
```diff
@@ -93,6 +94,50 @@ message MousebindOnPressRequest {
...
+  DOWN_AND_RIGHT = 5;
+  UP_AND_LEFT = 6;
+  UP_AND_RIGHT = 7;
+}
+
+enum GestureType {
+  HOLD = 0;
+  PINCH = 1;
+  SWIPE = 2;
+}
```
Ottatop: Enum names should follow the [protobuf style guide](https://protobuf.dev/programming-guides/style/#enums).

src/input.rs
```diff
@@ -1326,6 +1429,33 @@ fn constrain_point_inside_rects(
         .unwrap_or(pos)
 }
 
+fn delta_to_direction(delta: (f64, f64)) -> GestureDirection {
+    use std::f64::consts::PI;
+
+    let (x, y) = delta;
+
+    let angle = y.atan2(x);
+    let angle_deg = (angle * 180.0 / PI + 360.0) % 360.0;
```
Ottatop:
```suggestion
    let angle_deg = (angle.to_degrees() + 360.0) % 360.0;
```

src/api/input/v1.rs
```diff
@@ -548,6 +626,54 @@ impl input::v1::input_service_server::InputService for InputService {
...
+
+        run_unary_no_response(&self.sender, move |state| {
+            state
+                .pinnacle
+                .input_state
+                .bind_state
+                .gesturebinds
+                .set_gesturebind_has_on_begin(bind_id);
+        })
+        .await
```
Ottatop: Ok but why do we have a different rpc that does exactly the same thing as the one two above it?

api/protobuf/pinnacle/input/v1/input.proto
```diff
@@ -313,9 +358,13 @@ service InputService {
 
   rpc KeybindStream(KeybindStreamRequest) returns (stream KeybindStreamResponse);
   rpc MousebindStream(MousebindStreamRequest) returns (stream MousebindStreamResponse);
+  rpc GesturebindStream(GesturebindStreamRequest) returns (stream GesturebindStreamResponse);
 
   rpc KeybindOnPress(KeybindOnPressRequest) returns (google.protobuf.Empty);
   rpc MousebindOnPress(MousebindOnPressRequest) returns (google.protobuf.Empty);
+  rpc Gesturebind(GesturebindRequest) returns (google.protobuf.Empty);
+  rpc GesturebindOnBegin(GesturebindOnBeginRequest) returns (google.protobuf.Empty);
+  rpc GesturebindOnFinish(GesturebindOnFinishRequest) returns (google.protobuf.Empty);
```
Ottatop: What are `Gesturebind` and `GesturebindOnFinish` for?

Ottatop: New RPCs should return a custom message so we don't need to make a breaking change in case we ever need to return data.

api/protobuf/pinnacle/input/v1/input.proto
```diff
@@ -93,6 +94,50 @@ message MousebindOnPressRequest {
...
+
+enum GestureDirection {
+  DOWN = 0;
+  LEFT = 1;
+  RIGHT = 2;
+  UP = 3;
+  DOWN_AND_LEFT = 4;
+  DOWN_AND_RIGHT = 5;
+  UP_AND_LEFT = 6;
+  UP_AND_RIGHT = 7;
```
Ottatop: > (though maybe i'd personnally prefer UpLeft to UpAndLeft).

I agree with Ph4ntomas, X_AND_Y feels more like you need to do X first, then Y. I'd remove the ands here and elsewhere.

api/rust/src/input.rs
```diff
@@ -524,6 +538,145 @@ async fn new_mousebind_stream(
...
+            .callback_sender
+            .get_or_insert_with(|| new_gesturebind_stream(self.bind_id).block_on_tokio());
+        let _ = sender.send((Box::new(on_finish), Edge::Release));
+
+        Client::input()
+            .gesturebind_on_finish(GesturebindOnFinishRequest {
+                bind_id: self.bind_id,
+            })
+            .block_on_tokio()
+            .unwrap();
```
Ottatop, line 654: `keybind_on_press` is used to let the compositor know when to suppress any release events associated with the bind. I can see you've added `gesturebind_on_begin` which I believe is supposed to do the same thing, but there's no reason for this `gesturebind_on_finish` RPC.

src/input/bind.rs
```diff
@@ -525,3 +528,152 @@ impl Mousebinds {
...
+        };
+        gesturebind.borrow_mut().has_on_begin = true;
+    }
+
+    pub fn set_gesturebind_has_on_finish(&self, gesturebind_id: u32) {
+        let Some(gesturebind) = self.id_map.get(&gesturebind_id) else {
+            return;
+        };
+        gesturebind.borrow_mut().has_on_begin = false;
+    }
```
Ottatop, line 685: ?

src/api/input/v1.rs
```diff
@@ -235,6 +265,7 @@ impl input::v1::input_service_server::InputService for InputService {
             match input::v1::bind::Bind::Key(input::v1::Keybind::default()) {
                 input::v1::bind::Bind::Key(_) => (),
                 input::v1::bind::Bind::Mouse(_) => (),
+                input::v1::bind::Bind::Gesture(_) => (),
```
Ottatop, line 275: Need to add info for gesturebinds below

api/rust/src/input.rs
```diff
@@ -524,6 +538,145 @@ async fn new_mousebind_stream(
...
+}
+
+bind_impl!(Gesturebind);
+
+/// Creates a gesturebind on the [`DEFAULT`][BindLayer::DEFAULT] bind layer.
+pub fn gesturebind(
+    mods: Mod,
+    direction: GestureDirection,
+    fingers: u32,
+    gesture_type: GestureType,
```
Ottatop: Not the biggest fan of how `direction` gets ignored when `gesture_type` is pinch or hold. See overarching review comment for my thoughts.

Ottatop: You should create enums for `GestureDirection` and `GestureType` instead of exposing the generated protobuf enums to the public API.

src/api/input/v1.rs
```diff
@@ -160,6 +164,32 @@ impl input::v1::input_service_server::InputService for InputService {
 
                     bind_id
                 }
+                input::v1::bind::Bind::Gesture(gesturebind) => {
+                    let direction = GestureDirection::try_from(gesturebind.direction)
+                        .expect("invalid gesture direction value");
+                    let fingers = gesturebind.fingers;
+                    let gesture_type = GestureType::try_from(gesturebind.gesture_type)
+                        .expect("invalid gesture type");
```
Ottatop: Let's not panic here if the client does a no-no.

Ottatop (commented):
Non-exhaustive, preliminary review. I have not tested anything here.

Apart from the Lua API, you're also going to have to add gesture stuff in a few other places:
- https://github.com/pinnacle-comp/pinnacle/blob/3fe097eba26bc92dc83ba3ea8af8f6e4f6f9f661/src/input/bind.rs#L22-L87
- https://github.com/pinnacle-comp/pinnacle/pull/417/changes#r2744410970

There might be other places, I'm realizing that the way I've architected binds isn't great 😬.

I'm not a fan of having people pass in both a gesture direction and a gesture type. The direction is only used when the type is swipe. I would replace both of those with an enum, maybe something like
```rust
enum GestureType {
    Swipe(Direction),
    Pinch,
    Hold,
}
```

You're also going to have to suppress gesture events from reaching the client if we bind them. We should maybe look at what niri does. Also, gestures can be cancelled so we need to handle that as well.

src/input.rs
cassandracomar: just looking at the math, `(a + b) % b` should be equal to `a % b`, I'm pretty sure? moreover, when you convert to degrees, you should already be getting a value between 0 and 359 so you can elide the modulus as well.

Ottatop: > just looking at the math, `(a + b) % b` should be equal to `a % b`, I'm pretty sure?

Not when `a` is negative, which `atan2` will return for the lower two quadrants. The plus and modulus are for converting the result of `atan2` to the 0-360 degree angle around the x-axis so determining the direction is easier.

src/input/bind.rs
rustysec: this is left over from some scaffolding i was throwing in for firing a gesture when it starts vs when it ends... i'm still not sure that's worth pursuing at the moment.

Issue #236, opened 2024-05-09
Add window minimizing


Ottatop:
With the merging of #235 I think it finally makes sense to add the ability to minimize windows now.

ildar:
workaround: I configured "minimizing" as moving to a "special" tag. Not perfect but useful.

PR #420, opened 2026-01-29
Feature: window minimisation

sapient-cogbag:
I've made bare-minimum window minimising functionality (#236). Unsure if I've gotten to all the places I need to with it but preliminary testing shows that it seems to work at least somewhat ^.^. 

However - I'm uncertain if you might want to make it so that focusing on a window (at least in the API) unminimises it or at the very least refuses to focus it? Also there are some questions on how handling minimisation of not-yet-mapped windows should work (extra rules option? etc. - I just put it as a "not yet implemented" type message nya ^.^)

- sapient

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -167,9 +167,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.100"
+version = "1.0.101"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a23eb6b1614318a8071c9b2521f36b424b2c83db5eb3a0fead4a6c0809af6e61"
+checksum = "5f0e0fee31ef5ed1ba1316088939cea399010ed7731dba877ed44aeb407a75ea"
 dependencies = [
  "backtrace",
 ]
@@ -203,7 +203,7 @@ checksum = "0ae92a5119aa49cdbcf6b9f893fe4e1d98b04ccbf82ee0584ad948a44a734dea"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.114",
+ "syn 2.0.115",
 ]
 
 [[package]]
@@ -268,7 +268,7 @@ checksum = "9035ad2d096bed7955a320ee7e2230574d28fd3c3a0f186cbea1ff3c7eed5dbb"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.114",
+ "syn 2.0.115",
 ]
 
 [[package]]
@@ -325,9 +325,9 @@ dependencies = [
 
 [[package]]
 name = "avif-serialize"
-version = "0.8.6"
+version = "0.8.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "47c8fbc0f831f4519fe8b810b6a7a91410ec83031b8233f730a0480029f6a23f"
+checksum = "375082f007bd67184fb9c0374614b29f9aaa604ec301635f72338bb65386a53d"
 dependencies = [
  "arrayvec",
 ]
@@ -493,9 +493,9 @@ checksum = "5dd9dc738b7a8311c7ade152424974d8115f2cdad61e8dab8dac9f2362298510"
 
 [[package]]
 name = "bytemuck"
-version = "1.24.0"
+version = "1.25.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1fbdf580320f38b612e485521afda1ee26d10cc9884efaaa750d383e13e3c5f4"
+checksum = "c8efb64bd706a16a1bdde310ae86b351e4d21550d98d056f22f8a7f7a2183fec"
 dependencies = [
  "bytemuck_derive",
 ]
@@ -508,7 +508,7 @@ checksum = "f9abbd1bc6865053c427f7198e6af43bfdedc55ab791faed4fbd361d789575ff"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.114",
+ "syn 2.0.115",
 ]
 
 [[package]]
@@ -519,9 +519,9 @@ checksum = "8f1fe948ff07f4bd06c30984e69f5b4899c516a3ef74f34df92a2df2ab535495"
 
 [[package]]
 name = "bytes"
-version = "1.11.0"
+version = "1.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b35204fbdc0b3f4446b89fc1ac2cf84a8a68971995d0bf2e925ec7cd960f9cb3"
+checksum = "1e748733b7cbc798e1434b6ac524f0c1ff2ab456fe201501e6497c8417a4fc33"
 
 [[package]]
 name = "calloop"
@@ -609,9 +609,9 @@ dependencies = [
 
 [[package]]
 name = "cc"
-version = "1.2.53"
+version = "1.2.55"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "755d2fce177175ffca841e9a06afdb2c4ab0f593d53b4dee48147dfaade85932"
+checksum = "47b26a0954ae34af09b50f0de26458fa95369a0d478d8236d3f93082b219bd29"
 dependencies = [
  "find-msvc-tools",
  "jobserver",
@@ -627,9 +627,9 @@ checksum = "6d43a04d8753f35258c91f8ec639f792891f748a1edbd759cf1dcea3382ad83c"
 
 [[package]]
 name = "cfg-expr"
-version = "0.20.5"
+version = "0.20.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "21be0e1ce6cdb2ee7fff840f922fb04ead349e5cfb1e750b769132d44ce04720"
+checksum = "78cef5b5a1a6827c7322ae2a636368a573006b27cfa76c7ebd53e834daeaab6a"
 dependencies = [
  "smallvec",
  "target-lexicon",
@@ -658,10 +658,21 @@ dependencies = [
 ]
 
 [[package]]
-name = "clap"
-version = "4.5.54"
+name = "chacha20"
+version = "0.10.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c6e6ff9dcd79cff5cd969a17a545d79e84ab086e444102a591e288a8aa3ce394"
+checksum = "6f8d983286843e49675a4b7a2d174efe136dc93a18d69130dd18198a6c167601"
+dependencies = [
+ "cfg-if",
+ "cpufeatures 0.3.0",
+ "rand_core 0.10.0",
+]
+
+[[package]]
+name = "clap"
+version = "4.5.58"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "63be97961acde393029492ce0be7a1af7e323e6bae9511ebfac33751be5e6806"
 dependencies = [
  "clap_builder",
  "clap_derive",
@@ -669,9 +680,9 @@ dependencies = [
 
 [[package]]
 name = "clap_builder"
-version = "4.5.54"
+version = "4.5.58"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fa42cf4d2b7a41bc8f663a7cab4031ebafa1bf3875705bfaf8466dc60ab52c00"
+checksum = "7f13174bda5dfd69d7e947827e5af4b0f2f94a4a3ee92912fba07a66150f21e2"
 dependencies = [
  "anstream",
  "anstyle",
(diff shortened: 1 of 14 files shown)


src/api/window.rs
```diff
@@ -65,7 +65,43 @@ pub fn set_geometry(
...
+    let is_minimized = window.with_state(|state| state.minimized);
+    let set = match set {
+        Some(absolute_set) => absolute_set,
+        None => !is_minimized,
+    };
+
+    // Note: tag moving will automatically adjust the output on the window directly even if
+    // minimised, so we can rely on this.
+    let Some(output) = window.output(&state.pinnacle) else {
+        // No associated output, do nothing...
```
Ph4ntomas: Shouldn't we at least warn here and/or prevent un-minimizing the window ? I don't know what are the side effect for clients to be un-minimized without an output.

sapient-cogbag: I think warning might be a good idea, but if we're talking programmatic unminimising then someone might unminimise a window with a hidden tag before then switching to that same tag. When writing this I've been essentially trying to treat "minimisation" like having a special, "hidden" tag ^.^.

sapient-cogbag: Actually wait - just realised that's the case of the window having no output at all (whether active or not). Fair enough.

sapient-cogbag: It could also be possible to adjust it regardless of the presence of the output - idk which would be preferred tho >.<.

sapient-cogbag: Just implemented this.

api/rust/examples/default_config/main.rs
```diff
@@ -151,6 +151,16 @@ async fn config() {
         .group("Window")
         .description("Toggle maximized on the focused window");
 
+    // `mod_key + n` toggles minimized
+    input::keybind(mod_key, 'n')
+        .on_press(|| {
+            if let Some(window) = window::get_focused() {
+                window.toggle_minimized();
+            }
```
Ph4ntomas, line 159: I expect this will set the minimized flag, then the window to lose focus, so the next time would either have no window focused and fail, or another window.

I think what awesome did makes more sense. Minimized window are pushed on a stack,  and you have two keybinds: one to push a window on that stack and another to pop from that stack.

I don't know if the stack make sense, but IMO if the behavior is the one I mentioned we should split the keybinds as well, because the actual behavior is not to toggle the flag.

sapient-cogbag: This is how I always actually do it in awesomeWM tbh. I use a switcher (rofi for awesomewm, though it seems to have trouble doing percent-widths on pinnacle, but other tools like anyrun manage fine so I think it's a rofi issue nya) or the bar (not yet a thing in pinnacle unless you were to make one with snowcap and I've not messed with that yet).

Might be sensible to make it do `set_minimized` instead tho.

Ph4ntomas: > I think what awesome did makes more sense. Minimized window are pushed on a stack, and you have two keybinds: one to push a window on that stack and another to pop from that stack.

The stack thing might just be something from my config (regardless it's something that can be implemented config-side either way).

> not yet a thing in pinnacle unless you were to make one with snowcap and I've not messed with that yet

I have one, and #414 is a step in that direction :)

Regardless of the bar status tho, my point was more that since there's no way to focus a minimized window, a toggle keybind make less sense than having one binding to minimize the focus window, and another one to un-minimize the last minimized window.

cassandracomar: if fetching the windows in active tags still includes the minimized windows, you'd still be able to focus minimized windows by unminimizing the window, then moving focus.

Ph4ntomas: @cassandracomar sure I'm not saying you can't unminimized, just that you can't use `get_focused()` to get a minimized window, so this function can't toggle the state.

cassandracomar: ahh, yeah fair

sapient-cogbag: Not being able to have focus on a minimised window is at least theoretically intended to prevent confusion. I personally think it would make sense to have focusing a window automatically unminimise that window, but I'm uncertain of if it is preferred for the APIs to be disjoint (also while I was writing it I wasn't certain how the unmapped vs mapped windows worked since there is a function that reverts a mapped window to unmapped state without making it be deleted, so for a while I thought the mechanism for hiding windows was unmapping them, essentially because the "unmapped vector" exists and is distinct from windows that are unmapped, I _think_ nya).

I do agree that having an example of both minimisation and unminimisation would be a good idea though. I'd use the focus stack and find the first minimised window on the active tags, or something like that, as Cassandra suggested.

The general idea I had with minimisation however, is it's set up to preserve things like size (fullscreen or maximised) etc. so when unminimised the window "goes back to how it's expected" (also accounting for things like having tags added/removed, shuffling outputs etc.).

I'm also unsure if we should even be unfocusing when minimising, as the focus stack already filters out minimised windows when finding the focused window. However, I'm pretty sure this is what you want so that the user controls the next focus. In fact, not unminimizing windows when they are  focused allows you to "preemptively" focus a minimised window programmatically so that when it gets unminimised it has focus the moment it happens, but I'm pretty sure this is not worth the confusing behaviour of potentially allowing input to a minimised window nya.

As for obtaining lists of windows, that can absolutely be done in the config/client and includes minimised windows. I think I will modify the config as suggested, and also make it so that focusing a window unminimises it ^.^.

sapient-cogbag: Well, right now, focusing on a minimised window would focus a different window (enabling focus but then having the filter filter it out...). So auto-unminimising is the only sensible solution here (IMO ofc).

api/rust/examples/default_config/main.rs
```diff
@@ -161,6 +161,23 @@ async fn config() {
...
 
+    // `mon_key + shift + n` unminimises the "most recently focused"
+    // minimised window that is on the active tags in this current output.
+    input::keybind(mod_key | Mod::SHIFT, 'n')
+        .on_press(|| {
+            let Some(output) = output::get_focused() else { return };
+            let Some(most_recently_minimised_active_window) = output
+                .keyboard_focus_stack_visible()
+                .filter(|w| w.minimized())
+                .last()
```
sapient-cogbag, line 172: Note on recent modification: The name of the API function "keyboard_focus_stack_visible" seems misleading now that minimised windows are a thing >.<

api/protobuf/pinnacle/window/v1/window.proto
```diff
@@ -231,6 +244,7 @@ service WindowService {
   rpc ResizeTile(ResizeTileRequest) returns (google.protobuf.Empty);
   rpc SetFullscreen(SetFullscreenRequest) returns (google.protobuf.Empty);
   rpc SetMaximized(SetMaximizedRequest) returns (google.protobuf.Empty);
+  rpc SetMinimized(SetMinimizedRequest) returns (google.protobuf.Empty);
```
Ottatop: New RPCs should return a custom message so we don't have to add a new RPC if we want to return an actual response.

src/api/window/v1.rs
```diff
@@ -479,6 +500,31 @@ impl v1::window_service_server::WindowService for super::WindowService {
...
+        };
+
+        run_unary_no_response(&self.sender, move |state| {
+            if let Some(window) = window_id.window(&state.pinnacle) {
+                crate::api::window::set_minimized(state, &window, absolute_minimized);
+            } else if let Some(unmapped) = window_id.unmapped_window_mut(&mut state.pinnacle)
+                && let UnmappedState::WaitingForRules { rules: _, .. } = &mut unmapped.state
+            {
+                // TODO: find a way to immediately minimize a window upon mapping.
+                warn!("minimizing unmapped windows not yet supported");
```
Ottatop: Does adding `minimized` to the `WindowRules` [here](https://github.com/pinnacle-comp/pinnacle/blob/0a34a73b4db290748667378dfc4f44e2a8323206/src/window/rules.rs#L35) and applying that [here](https://github.com/pinnacle-comp/pinnacle/blob/0a34a73b4db290748667378dfc4f44e2a8323206/src/window/rules.rs#L151) work?

api/rust/examples/default_config/main.rs
Ottatop: Hm guess we can deprecate it and add a renamed function

src/api/window/v1.rs
sapient-cogbag: It does seem to, with my preliminary testing.

However during testing I discovered that getting the focus stack of active tags no longer includes minimised windows, since they seem to lose their location value (which makes sense, but I'm not certain what exactly to do about it). During testing I made a custom config that retrieved all windows then filtered them for windows on the focused output and also minimization, but it is less convenient (but still doable if you combined it with some sort of active tag filter).

api/rust/examples/default_config/main.rs
sapient-cogbag: Some sort of change seems to have made minimized windows not show up in it anyway (as they now lack a location), which makes more sense for the name anyway IMO (as they're not visible :p). But this also breaks the focus stack as a more general tool since it filters out windows by location down in the element check and minimization deletes that for some reason. Or at the very least the minimized windows don't get listed for their associated outputs and I think it's for that reason based on some testing >.<...

Ph4ntomas:
> I'm not sure having "arbitrary retrieval" with underspecified behaviour would necessarily be the best idea, personally, but that's just my random opinion ^.^ nya

Well 'underspecifying' here is mostly a way have something that work without needing to maintain a behavior or giving too many informations about the internals, but we don't have to duplicate AwesomeWM behavior on everything either. 

Window order is already unspecified in other part of the API since pinnacle reorder window when you move them around and don't sort them in get() functions, so I don't really see a point in enforcing an order for minimized window. However, I would prefer a function returning a list of minimized window with unspecified ordering than a function un-minimizing a random window and returning its handle.

> Might need some help with deprecating functions properly (also in how to change the functions properly), I'm not entirely sure about the versioning of the API crate and the protocols and how they interact, though my best guess is that the protocol is basically "v1 for the forseeable future" but the API crate policy is as described in the wiki.

#408 has an example of deprecating messages. I think RPCs support similar annotation, but I'm not sure. As for versioning, my understanding is that the server must support any client that use the same "version namespace". However, we can add new RPC, or even deprecate some RPCs, as long as we keep serving the requests *in some way*. 

As an example, if a client calls `SetFocus` on a minimized window, the RPC would fail silently (i.e. succeed but without actually setting the focus on minimized window). If the client support `SetFocusV2` (my 2c is that `TrySetFocus` is a better name in this case), then it will receive the error.
For consistency, the client-side API should have the WindowHandle method marked as deprecated so user know they should adapt their code, and the server should warn that a deprecated RPC was called so a user can find the reason the RPC had no side-effect if using an older client.

api/rust/src/window.rs
```diff
@@ -235,6 +235,29 @@ impl VrrDemand {
...
+    fn try_from(
+        value: pinnacle_api_defs::pinnacle::window::v1::TrySetFocusedResponse,
+    ) -> Result<Self, Self::Error> {
+        use pinnacle_api_defs::pinnacle::window::v1::try_set_focused_response::TrySetFocusedStatus;
+        match value.status() {
+            TrySetFocusedStatus::Success => Err(()),
+            TrySetFocusedStatus::WindowMinimized => Ok(Self::WindowMinimized),
+        }
+    }
+}
```
Ph4ntomas: Just my 2c, but this feels weird (as does the whole swap_ok_err() extension method). 

Is there a reason not to implement `From<TrySetFocuseResponse>` for `Result<(), TrySetFocusedError>` instead ?

Ph4ntomas: Since it's just a gut-feeling, maybe don't act on it until Ottatop give an input :)

If the issue is that rust don't like when we implements foreign traits on foreign types, it was solved on the server-side by having a `FromApi` trait, with the same semantics as `From`, and I feel it make the code a bit cleaner

sapient-cogbag: Ahhh... I probably should have done that instead of TryFrom, yes. There was a reason (when I had TrySetFocusedError as part of the protobuf generated rust files, I couldn't implement the trait due to orphan rules) but it shouldn't be an issue now ^.^

sapient-cogbag: Actually, scratch that, it still doesn't work. I could try implementing `Into` instead, but that's really not ideal nya.

Edit: this also doesn't work. Didn't think it would but just wanted to check. The problem is that both `Result` and the `TrySetFocusResponse` type are foreign to the main pinnacle crate.

sapient-cogbag: We could also create a separate `TryFrom`/`TryInto`-like pair of traits instead for responses with embedded errors and use that.

Ph4ntomas: > Actually, scratch that, it still doesn't work. I could try implementing Into instead, but that's really not ideal nya.
> 
> Edit: this also doesn't work. Didn't think it would but just wanted to check. The problem is that both Result and the TrySetFocusResponse type are foreign to the main pinnacle crate.

I'm not sure what you tried, but the orphan rule shouldn't apply if the trait is local to the crate (i.e. if you copy https://github.com/pinnacle-comp/pinnacle/blob/main/snowcap/src/util/convert.rs to from server-side to the config api & use FromAPI instead of From)

sapient-cogbag: Yes. I was using the "default" `(Try)Into`/`(Try)From` traits from stdlib. It would work if I used a trait local to the crate. I will do this.

PR #422, opened 2026-02-03
Snowcap: Add support for touch input

Ph4ntomas:
This adds support for touch input for snowcap.

When a touch input is received on a surface, it's converted to an iced touch input event, and sent to the surface widget.

Iced currently ignores position when handling touch inputs. As a workaround, I'm updating the surfaces cursor position when processing input. This isn't ideal because the cursor position and the mouse gets temporarily disconnected, but is safe to do because the last known position is always sent in pointer events. This however prevent mouse_area from firing "Leave" events, because the position will only be updated on the next input (mouse or touch) event, but I felt it was better than missing the release event.

Hopefully, https://github.com/iced-rs/iced/issues/3234 will solve this, and improve touch inputs in general at some point.

---------------------
I'm opening this up for review, but there is a few thing that still need to be done:

- [x] better documentation (example, lua-doc, ...)
- [x] some decision on the improved button & mouse-area. Should they be kept in-tree ? Should they be out of tree ? Not included ? (even tho they are broken for touch control, the toucharea works so the patched version are not mandatory)

The name also somewhat became misleading as this PR no longer only make changes in snowcap, For the full changes:
- touching a window will set the focus to that window.
- MoveSurfaceGrab and ResizeSurfaceGrab (interactive move/resize) can now support touch-based inputs.
- It's possible to start touch-based grab from the config.
- Touch area widget can be used to retrieve finger_id as well as touch related events. This allows for example to start a resize-grab when touching the border of a floating window.

### DIFF
--- a/api/rust/src/window.rs
+++ b/api/rust/src/window.rs
@@ -24,7 +24,7 @@ use pinnacle_api_defs::pinnacle::{
             MoveToTagRequest, RaiseRequest, ResizeGrabRequest, ResizeTileRequest,
             SetDecorationModeRequest, SetFloatingRequest, SetFocusedRequest, SetFullscreenRequest,
             SetGeometryRequest, SetMaximizedRequest, SetTagRequest, SetTagsRequest,
-            SetVrrDemandRequest, SwapRequest,
+            SetVrrDemandRequest, SwapRequest, TouchMoveGrabRequest, TouchResizeGrabRequest,
         },
     },
 };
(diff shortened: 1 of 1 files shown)


Ph4ntomas:
Touch "bind" are left out of scope on purpose. There are a bunch of stuff I need to think about before implementing them.

I'd also like another type of grab that would make use of the multi-touch capabilities of touch-screen, which should also be kept for later.

PR #424, opened 2026-02-05
snowcap: correct popup destruction on decoration close

Ph4ntomas:
This is about as tested as popup on decoration, but I had some crashed during popups development when I did not explicitly manage destruction order, so let's not do that :)


Ottatop (approved):
Thanks!

PR #425, opened 2026-02-07
api: provide an error when the output does not exist where to add an tag

xMAC94x:
similar to the error for `move_to_output`

### DIFF
--- a/api/rust/src/tag.rs
+++ b/api/rust/src/tag.rs
@@ -28,7 +28,7 @@ use pinnacle_api_defs::pinnacle::{
     tag::v1::{
         AddRequest, GetActiveRequest, GetNameRequest, GetOutputNameRequest, GetRequest,
         MoveToOutputRequest, RemoveRequest, SetActiveRequest, SwitchToRequest,
-        move_to_output_response::error::Kind,
+        add_response::error::Kind as AddErrorKind, move_to_output_response::error::Kind,
     },
     util::v1::SetOrToggle,
 };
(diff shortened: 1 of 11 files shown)


api/rust/src/tag.rs
```diff
@@ -53,29 +53,37 @@ use crate::{
...
-        .into_inner()
-        .tag_ids
-        .into_iter()
-        .map(|id| TagHandle { id })
+        .into_inner();
+
+    let error = response.error.and_then(|error| error.kind);
+    match error {
+        None => Ok(response.tag_ids.into_iter().map(|id| TagHandle { id })),
+        Some(AddErrorKind::OutputDoesNotExist(_)) => Err(AddError::OutputDoesNotExist),
```
xMAC94x, line 85: It feels kind of weird. The api gets a `OutputHandle` and we can prob always expect the outputHandle to actually map to an existing Output.
It feels weird that you have to check on that error as user again. Maybe its okay to just `tracing::error()` or expect in this case.

The same goes for `move_to_output` when the output is provided via a handle.

WDYT ?

Ph4ntomas: > The api gets a OutputHandle and we can prob always expect the OutputHandle to actually map to an existing Output.

The OutputHandle could be stale or become stale sometime between calling this function and the server receiving the message. We can't do any assumption on the validity of the data (beside, user input should not be trusted, so let's not assume anything).

> It feels weird that you have to check on that error as user again. Maybe its okay to just tracing::error() or expect in this case.

IMO the API should never hide or consume errors, unless the "error" is an expected case and a documented fallback behavior exists (and even then the calling code should be notified).

xMAC94x: in that case I think the PR is good to go

Ph4ntomas:
> After giving this some thought, my main concern is that returning an error introduces a bit more friction when using this API. Yes, if we really wanted to be "correct" here we would add this. However, there are also a slew of methods on the other *Handles that also silently fail when the associated object is gone, and making those also return errors would kinda suck tbh. I'm not sure where to draw the line.

Just my 2c here. Most of the time, errors will end up in the log file by the server. This is more that okay for now, but IMO it's confusing for user not to have any kind of (visual) feedback. OTOH keeping the API simple is nice.

My proposition would be:
- We transmit errors to the client/config via protobuf.
- The current API is kept as-is¹.
- We add `try*` method if users want to perform error checking themselves. 

¹ At some point, we could log errors/warning coming from the server in non-failible function by calling `tracing::*` method. This would allow user to add a sink that would transform the errors in some kind of notification and display. The fallback sink could just be a println as we currently have.

Ottatop:
Let's hold off on that until we figure out how we're actually going to *use* the errors, no point in implementing that now if plans change later

Issue #427, opened 2026-02-08
[Snowcap] Optional View for widgets


Ph4ntomas:
While porting Glacier's Bar to the new composable Program API, I stumbled upon a slight issue:

Some widgets are meant to be hidden until enabled (e.g. a prompt). I used to do so by having my own `Program` equivalent where the view was Optional, and filtering them out in the parent Widget/the toplevel Program.

I see two way of fixing that:
- Make `Program::view` return an Option (IMO cleaner)
- Make a special Widget alternative that represent an empty widget.

In both case, the None/Empty value should never reach the server (imo). The goal is to remove them from the view when not needed.

An alternative is to use an empty container (e.g. row or column), but this is a bad habit. I'd much rather have a Program return a value we can handle on the client side, since sending en empty row or column *can* crash snowcap (see #404)

Ph4ntomas:
A few Pros & Cons of returning an Option :
If Program::view returns an Option we can make use of rust Iterator to automatically remove the empty widget.
Preventing the option from crossing the client boundaries is a simple `.expect("")` in the `Surface` code if the user code does a no-no (but we can handle it further).
It *is* technically a breaking change, but the discussion in #408 seemed to assume that snowcap was to be considered *very* unstable.

Ph4ntomas:
> I just realized, we could just return an empty row if we get None for a toplevel program. That makes making view optional much more palatable.

That's an option, anything with at least 1x1 area can work (or an empty row once #404  is fixed). TBH after playing more with this (hence #429), I think I would prefer a clearer separation between program able to act as toplevel (for those it makes sens to enforce some widget to be returned) and programs meant only as hosted widgets.

PR #428, opened 2026-02-08
snowcap: Make Program::view return an option

Ph4ntomas:
Snowcap `Program` are now meant to both act as top-level handles and as composable fragment of the widget tree. Having the `view()` function return an option allow parent widget to filter-out empty views, instead of having the child needlessly return an empty container.

If this cannot be merged as-is due to breaking the API, we could solve this using a sub-trait `SubProgram`, with a blanket implementation for `Program` and a `try_view` function instead of the view function. We'd also need to change `Program::register_child` to take `SubProgram` instead of `Program`.

However, I don't know if I like having that distinction, so I changed the `Program` trait instead.

Fixes: #427

### DIFF
--- a/api/rust/src/experimental.rs
+++ b/api/rust/src/experimental.rs
@@ -22,11 +22,13 @@ pub mod input_grab {
 
         fn update(&mut self, _msg: Self::Message) {}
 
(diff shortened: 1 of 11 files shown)


Ph4ntomas:
Taking this back to draft. It work for my use-case but causes other issues I'd prefer not to have down the line.

PR #430, opened 2026-02-12
429 composable program v2

### LINKED ISSUE #429: [Snowcap] Composable Program feedback
I've been busy porting Glacier to the composable programs. Here are some thought/feedback from using it more than the toy tests during the Review, before pushing for upstreaming stuff.

I have some auto-hiding widgets (e.g. the prompt, various text box). I'd much prefer if they were able to return `None` when there's nothing to view. (I've open #427 then #428, but I'll present some thought after the overall feedback)

Right now, the plugins can opt-in to have full, constant access to the handle. This means they have 2 channels to send messages, (through signals and through the handle messages). It also causes some issues with separation of responsibility. 
As an example, when activating a prompt, the widget need to be focused and the layer need to get keyboard interactivity. In my previous iteration, there was a separate `Widget` trait, which acted as a program-light which did not keep access to the surface outside of the update function (and even then it was only used for popup). This meant the prompt needed to forward the focus request to the `Program` hosting it, which then used the surface handle to focus & take keyboard interactivity. With the new design, both the 'toplevel' Program and any of it's subprogram may act on the handle to change focus & interactivity, and there's nothing to prevent that.

The UniversalMessage object & Universal trait could be improved a bit for use in toplevel program. I have a few messages handled both by the toplevel Program & the widget (e.g. for focus management), meaning my bar need to clone every universal messages to probe whether it might need to be handled or not. IMO we either need a way to test the message type (`is<T>() -> bool`), or to access the message content as a reference (`downcast_ref<T>() -> Option<&T>`)

It might be because I haven't had enough time to play with it but I feel having the same trait for both 'toplevel' program that are tightly integrated with the surface they're on and 'hosted' may lead to some issue down the line with incompatible 'hosted programs' because they expect the toplevel to act a certain way. 

Some closing thought:

IMO we could improve some of this by having a `Program`-like trait `hosted::Program`, with the following changes:
- `SurfaceHandle` would be replaced by a more restrictive/weak type (`hosted::SurfaceHandle`), so the `hosted::Program` can't just access the raw handle.
- some of `hosted::SurfaceHandle` methods wouldn't be applied to the underlying handle, but instead bubbled-up to the hosting `Program` (e.g. `operate`, `close`). Less critical ones may be kept as-is (e.g. `send_message`, `force_redraw`). (Note: critical methods like operate & close could be replaced by signal, too).
- `hosted::Program::view` doesn't have to return a `WidgetDef` (instead of merging #428).
- `hosted::Program` can be composed from other `hosted::Program` only. `Program` may decide to forward call to the `Program` trait to other `Program`, and should otherwise properly call `hosted::Program` functions.
- `hosted::Program` are surface-agnostics and `Program` agnostic, while `Program` may access the raw Handle that was passed during the `created` call.
- standard `hosted::Program` are expected to use `UniversalMessages`. Standard `Program` should if they are also `hosted::Program`.
- standard `Program` may also define their own `Message` type for when they're not hosted. Their custom message type must be convertible from `Universal`. Example use case: a Bar may an hypotetical "HideBar" which it can only use in hosting mode. When hosted, it must abide by the hosting `Program` Message, meaning it can't make use of it. Regardless, it must handle the `UniversalMessage` to forward them to its children.

This proposal shouldn't hurt composability, as `Program` could also implement `hosted::Program` and re-use most of the code but wouldn't mess with the toplevel `Program` expectations when hosted. Host-only program can be more tightly integrated with their surface and ensure correct behavior (e.g. focus-arbitration if the user hit `tab` or escape, correct order of operation if a widget discard focus while another-one has it, etc.). Hosted-only `Program` can be written with some expectation about the behavior of the host `Program`.

@Ottatop Sorry I did not spot these issue during the review. Let me know if you're interested in some POC implementation.

Ph4ntomas:
Okay, this implements the latest state of the discussion in #429.

CHANGES:
- implements Signal on widget::operation::Operation.
- surface now manage their keyboard exclusivity. Layer does that when plugins requests to be focused/unfocused.
  I did not implement that when the operation comes from outside the window, only for signals.
  Right now there's no way for a program to request keyboard exclusivity without requesting focus on a widget at the same time. It can be done from outside the program tree.
- Operation can only be sent via signal.
- a new RequestClose signal was added. It replace the close methode on the surface handler
- `Program::view` not return an optional. If it bubble up to the Surface, the surface replace it by an empty list.

I left send messages & force_redraw because they are less problematic. They *can* be replaced by signal, too.

There is one thing "bothering" me:
I'd like Programs to be notified when the surface drop focus/keyboard interactivity via a message, but there's not way to forward such messages since surfaces are generic over their `Message` type. I wonder if we should have a curated list of "StandardMessage" and force Program::Message to implement "From<StandardMessage>" (but tbh I can leave without it)

~~I still need to implements the lua side.~~ Lua side is done.

supersedes: #428 
closes #429

### DIFF
--- a/snowcap/api/rust/src/surface.rs
+++ b/snowcap/api/rust/src/surface.rs
@@ -11,7 +11,7 @@ pub mod layer;
 pub mod popup;
 
 /// Events emitted by the surface to notify [`Program`] of state changes.
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 #[non_exhaustive]
 pub enum SurfaceEvent<Msg> {
     /// Emitted when the surface is created.
@@ -34,6 +34,19 @@ pub enum SurfaceEvent<Msg> {
     FocusGained,
     /// Emitted when the surface loses focus.
     FocusLost,
+}
+
+impl<Msg> Clone for SurfaceEvent<Msg> {
+    fn clone(&self) -> Self {
+        match self {
+            Self::Created { surface } => Self::Created {
+                surface: surface.clone(),
+            },
+            Self::Closing => Self::Closing,
+            Self::FocusGained => Self::FocusGained,
+            Self::FocusLost => Self::FocusLost,
+        }
+    }
 }
 
 /// Implementation detail for [`SurfaceHandle`]


snowcap/api/rust/src/surface/layer.rs
```diff
@@ -426,6 +458,19 @@ where
             on_press(handle, event.key, event.mods)
         });
     }
+
+    /// Manage keyboard interactivity on focus change.
+    fn handle_focus(&self, operation: &focusable::Focusable) {
+        match operation {
+            focusable::Focusable::Unfocus => {
+                let _ = self.set_keyboard_interactivity(KeyboardInteractivity::None);
+                self.send_message(surface::message::FocusYielded.into_universal().into());
```
Ph4ntomas: I'm doing this here for now, but I think I'd like to have that message come from the server.

The way I see it, I'd like the following informations at least:
- FocusGained: self-explicit
- FocusLost: The focus is temporarily lost and may be regained automatically (e.g. a child surface was opened)
- FocusYielded: The focus is lost and will not be regained automatically (e.g. we dropped keyboard interactivity).
- Closing: The surface is closing. This is the last chance to emit a signal to inform other part of the config that we're about to close. This is useful for e.g. popup that are being discarded by the compositor, so their parent layer can update their state.

snowcap/api/rust/src/surface/layer.rs
```diff
@@ -426,6 +458,19 @@ where
             on_press(handle, event.key, event.mods)
         });
     }
+
+    /// Manage keyboard interactivity on focus change.
+    fn handle_focus(&self, operation: &focusable::Focusable) {
+        match operation {
+            focusable::Focusable::Unfocus => {
+                let _ = self.set_keyboard_interactivity(KeyboardInteractivity::None);
```
Ph4ntomas: I'm wondering if we should have a `KeyboardInteractivity::Previous` or `KeyboardInteractivity::Default` here, instead of None. This would allow the layer to have On-demand interactivity (i.e. on click) by default, but grab interactivity programmatically.

Previous/Default would let the server know that we want to come back to the state used on creation.

This can come later tho, but just a thought.

snowcap/api/protobuf/snowcap/layer/v1/layer.proto
```diff
@@ -71,10 +71,32 @@ message ViewRequest {
...
+  LayerEvent layer_events = 1;
+}
+
 service LayerService {
   rpc NewLayer(NewLayerRequest) returns (NewLayerResponse);
   rpc Close(CloseRequest) returns (google.protobuf.Empty);
   rpc OperateLayer(OperateLayerRequest) returns (OperateLayerResponse);
   rpc UpdateLayer(UpdateLayerRequest) returns (UpdateLayerResponse);
   rpc RequestView(ViewRequest) returns (ViewResponse);
+  rpc GetLayerEvents(GetLayerEventsRequest) returns (stream GetLayerEventsResponse);
```
Ph4ntomas, line 100: I'm unsure whether this should be a layer-only event or a more generic surface event.

I've put it there because `FOCUS_YIELDED` makes no sense for a popup, and none of the focus events are currently relevant for decorations.

Ph4ntomas: Okay, I give it a bit more thought. IMO it's more future-proof to have surface event per surface type.

Maybe a `surface.proto` could contain shared types tho

Ottatop: Ye that's fine with me.

> Maybe a `surface.proto` could contain shared types tho

I think it's better to keep the types separate in case we need to alter the type for a specific surface.

Ph4ntomas:
> I'm not particularly attached to the name, but I'd like to have the ability to distinguish between those events. (If not, we'll have to do more manual focus management, that's fine too, but it'll be harder)

TBH I could just rework Glacier's to call the `Prompt`'s handle `deactivate` method in key handler callback (or send a message after requesting the keyboard interactivity change, so we can drop it if you don't want that

snowcap/api/rust/src/surface/layer.rs
```diff
@@ -202,12 +211,20 @@ where
...
-        msg_sender: msg_send,
-    };
+        signaler.connect({
+            let handle = handle.clone();
+
+            move |operation: operation::Operation| {
+                if let operation::Operation::Focusable(f) = &operation {
+                    handle.handle_focus(f);
+                };
+
```
Ph4ntomas: I think I should fold that in the handle `operate` method. I don't think it's a good idea to have the behavior differ depending on whether it's coming from a signal or a direct call.

Ph4ntomas: Done

Ph4ntomas:
Alright, I think that should be it

Ph4ntomas:
> Alright, I think that should be it

Well I lied :D

I've also removed `send_message` from the opaque `SurfaceHandle`. Messages can already go through signals, I don't really see a good reason in having a second channel.

FWIW, we could do it the other way around too (drop the message, but keep the method) but going through the signal chain has a few advantages:
- a program that doesn't otherwise need the handle don't have to keep it around
- other element of the config may listen to some messages (e.g. listen to some program `Done` message).

I'm leaving in `force_redraw` as IMO there shouldn't be any way to filter those out, and I don't really see how useful listening to them would be. 

I'm keeping `RedrawNeeded` signal since some widget may not care about the SurfaceHandle but still want to trigger redraws, or the other way around (they may not need to send signals except `RedrawNeeded`, so they don't have a signaler.)

snowcap/src/layer.rs
```diff
@@ -70,12 +78,15 @@ pub struct SnowcapLayer {
     // COMPAT: 0.1
     max_size: Option<iced::Size<u32>>,
 
+    keyboard_interactivity: wlr_layer::KeyboardInteractivity,
```
Ottatop: This is unused

snowcap/src/handlers/keyboard.rs
Ottatop, line 146: Also need to handle popup focus changing

snowcap/api/rust/src/surface/layer.rs
```diff
@@ -378,6 +473,16 @@ impl<Msg> LayerHandle<Msg> {
...
+    /// Manage keyboard interactivity on focus change.
+    fn handle_focus(&self, operation: &focusable::Focusable) {
+        let _ = match operation {
+            focusable::Focusable::Unfocus => {
+                self.set_keyboard_interactivity(KeyboardInteractivity::None)
+            }
+            _ => self.set_keyboard_interactivity(KeyboardInteractivity::Exclusive),
+        };
+    }
```
Ottatop: Won't this remove `OnDemand` interactivity on unfocus?

snowcap/api/protobuf/snowcap/layer/v1/layer.proto
```diff
@@ -71,10 +71,31 @@ message ViewRequest {
...
+    FOCUS_LOST = 2;
+  }
+
+  oneof event {
+    Focus focus = 1;
+  }
+}
+
+message GetLayerEventsResponse {
+  LayerEvent layer_events = 1;
```
Ottatop: Nit
```suggestion
  LayerEvent layer_event = 1;
```

snowcap/api/protobuf/snowcap/popup/v1/popup.proto
```diff
@@ -108,10 +108,31 @@ message ViewRequest {
...
+    FOCUS_LOST = 2;
+  }
+
+  oneof event {
+    Focus focus = 1;
+  }
+}
+
+message GetPopupEventsResponse {
+  PopupEvent popup_events = 1;
```
Ottatop: Nit
```suggestion
  PopupEvent popup_event = 1;
```

snowcap/api/protobuf/snowcap/layer/v1/layer.proto
Ph4ntomas: I agree with the nit, but I was wondering whether this should be a `repeated` field. We don't need that for the time being, but we had the issue with widgets event.

But, we can always keep a `layer_event` now and handle the issue later, if it appears

snowcap/api/rust/src/surface/layer.rs
Ph4ntomas: It will. My plan was to add a fake "keyboard interactivity": `default` that would go back interactivity set on layer creation, but I must've drop that commit since the only remnant is an unused parameter ...

snowcap/api/protobuf/snowcap/layer/v1/layer.proto
Ottatop: Ah yea, that's a good point. I suppose there's no reason *not* to make it repeated, so let's do that.

snowcap/src/handlers/keyboard.rs
Ph4ntomas: whoops thanks !

snowcap/api/protobuf/snowcap/popup/v1/popup.proto
Ph4ntomas: Resolved as per the equivalent layer code (make it repeated)

snowcap/api/rust/src/surface/layer.rs
Ph4ntomas: I've re-added that :)

snowcap/api/lua/snowcap/layer.lua
```diff
@@ -266,16 +335,35 @@ end
...
+    if focusable.unfocus ~= nil then
+        self:update({
+            keyboard_interactivity = keyboard_interactivity.DEFAULT,
+        })
+    else
+        self:update({
+            keyboard_interactivity = keyboard_interactivity.EXCLUSIVE,
+        })
+    end
+end
```
Ottatop: Ok I'm beginning to think that focus operations messing with the keyboard interactivity isn't ideal because I'm realizing that there's an issue in the opposite direction: what if people want to focus something without forcing this to grab the keyboard?

It really seems like we're abusing this to create a grab mechanism which, yea there's not really any other way to do it, but at the very least I'd prefer to break that behavior off into something separate like the `RequestGrabFocus/RequestUngrabFocus` signals I mentioned [here](https://github.com/pinnacle-comp/pinnacle/issues/429#issuecomment-3875221135). Or maybe even don't mess with the keyboard interactivity at all and open anything you want a keyboard grab on in a popup? idk

Ph4ntomas: > what if people want to focus something without forcing this to grab the keyboard?

I don't see the point here. Focus in iced means widgets "Keyboard Interactivity". I'm not sure what use there would be in giving keyboard interactivity to another widget without the surface taking interactivity.

> open anything you want a keyboard grab on in a popup? idk

That wouldn't work since popup can't have keyboard interactivity unless their parent surface have it.

Anyway, I'll remove the keyboard interactivity handling for now. I still think there's a use-case, *but* we can push that to the user code which should have access to the surface.

Ph4ntomas: > I still think there's a use-case, but we can push that to the user code which should have access to the surface.

Just to go into a bit more detail for my reasoning: The code here is rather "low-level" and acts as building blocks. The keyboard interactivity can be dealt with in objects wrapping the surface (e.g. a Bar) or set by users in their callbacks, so if there´s any doubt as to usefulness/potential issues I'd rather not have it done automatically for the time being.

I've left the "Default" keyboard interactivity enum because it's still useful.

I said I think there's a use-case because it might not be obvious to new users that you *need* to explicitly request it if the surface doesn't have it, but OTOH writing your own toplevel `Program` is a bit more advanced, so I'm fine with it too

Ottatop: > Focus in iced means widgets "Keyboard Interactivity". 

I think this conflates iced focus with actual keyboard focus. Focus in iced is *widget focus*, irrespective of the focus state of the surface those widgets are on. It determines what keyboard input is captured by what widget when the surface *has* keyboard focus, but handling of that keyboard focus happens at the compositor level.

> I'm not sure what use there would be in giving keyboard interactivity to another widget without the surface taking interactivity.

Consider a media player on a focused, on-demand layer. It has a search bar where you type in `/` to start searching for tracks. When that happens, you'll want to send a `Focus` operation to focus that search bar, but with the auto-focus handling impl this will set exclusive interactivity. If you click off the search bar, now the player is stuck with exclusive interactivity; this *requires* an `Unfocus` operation to "ungrab", which is not what you want.

> > open anything you want a keyboard grab on in a popup? idk
> 
> That wouldn't work since popup can't have keyboard interactivity unless their parent surface have it.

Are you sure? The protocol states "The parent of a grabbing popup must either be an xdg_toplevel surface or another xdg_popup with an explicit grab" i.e. nothing requires the toplevel surface to have keyboard focus, only a parent popup. And indeed Waybar opens grabbing popups that grab the keyboard even with `None` keyboard interactivity.

> I said I think there's a use-case because it might not be obvious to new users that you *need* to explicitly request it if the surface doesn't have it, but OTOH writing your own toplevel `Program` is a bit more advanced, so I'm fine with it too

Yea I think documentation would serve us well here

Ottatop:
> I've left the "Default" keyboard interactivity enum because it's still useful.

Only the caller has access to setting `Default` now, right? I don't see how it's useful in this case when the caller knows what keyboard interactivity they set.

snowcap/api/lua/snowcap/layer.lua
Ph4ntomas: > Are you sure? [...]
> And indeed Waybar opens grabbing popups that grab the keyboard even with None keyboard interactivity.

Ah right, the toplevel parents don't need interactivity, but you need a serial from a user-event. If the popup is open via a click, you don't need keyboard interactivity because you have that event.

Ph4ntomas:
> Only the caller has access to setting Default now, right? I don't see how it's useful in this case when the caller knows what keyboard interactivity they set.

Anything that can access the surface handle. I guess we can remove it and keep that info config-side, too

EDIT: I had in mind the ability to have object temporarily hijack the keyboard exclusivity, then put it back to default, but it would work by explicitely passing the default too

snowcap/src/layer.rs
Ottatop: This is still unused

Ph4ntomas: Whoops, sorry it should be ok now

PR #434, opened 2026-02-15
Snowcap: Add derive macro for Signal and Universal

Ph4ntomas:
This PR adds derive macro for `Signal` and `Universal`.

The `Signal` macro add bounds `Clone + 'static` automatically on generic types. I did not implement any validation/pruning yet, meaning deriving signal for :
```rust
struct<Msg: Clone> AAA {
   [...]
}
```

Will lead to the following code being generated:
```rust
impl<Msg: Clone + Clone + 'static> Signal for AAA<Msg> {}
```

Which I don't quite like.

Draft until #430 is merged.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4566,6 +4566,7 @@ dependencies = [
  "futures",
  "hyper-util",
  "snowcap-api-defs",
+ "snowcap-derive",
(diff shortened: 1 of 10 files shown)


Ottatop (approved):
Thanks!

PR #445, opened 2026-03-30
Fix surface closing event

Ph4ntomas:
For some reason I assumed widget event stream would closed when the associated sender is droped.

Since it's not the case, we need to send the closing event when the surface is dropped by server.

### DIFF
--- a/api/lua/build/grpc_client.lua
+++ b/api/lua/build/grpc_client.lua
@@ -314,7 +314,7 @@ end
 ---
 ---@param request_specifier grpc_client.RequestSpecifier
 ---@param data table The message to send. This should be in the structure of `request_specifier.request`.
----@param callback fun(response: table) A callback that will be run with every response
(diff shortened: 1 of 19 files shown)


Ph4ntomas:
This is currently rebased on top of #444, but I'll rebase it properly later and remove the draft.

Ph4ntomas:
This works well, but I wonder if I should expand on the lua side of things.

As far as I can tell, the widget stream is kept open indefinitely.

OTOH, this is ready to go, so maybe looking into that can be pushed to a later date.

Ph4ntomas:
Put that back to draft as I want to improve things on lua-side of things. The current solution is a hack and I'm working on some better way to do so (namely returning the streams so they can be shutdown cleanly when the surfaces are closed)

Ph4ntomas:
Closing this PR. Merged into #444

PR #446, opened 2026-04-05
snowcap: lua: Expose base.signaler

Ph4ntomas:
This PR exposes the signaler from base.

I have a Clock Program that share its signaler with the timer driving it, but I can't port it because the lua's `base.signaler` is private. (well I can, but it makes luals sad).

In rust, widget can contain a `Base` object and access that object's signaler.
In lua, all `Programs` inherit from `Base`, but the signaler is private.

This PR re-align lua with rust by adding a `base:signaler()` method, so plugins can share the signaler as a whole if needed.


Ottatop (approved):
Thanks!

PR #449, opened 2026-04-27
Fix clippy 0.1.97

Ph4ntomas:
Fixes the following new warning/clippy lint:

Latest rust environment added new warning for float -> f32 coercion as this will become a hard error in the future.
https://github.com/rust-lang/rust/issues/154024

Clippy now complains if an explicit index is used with iterator instead of `enumerate` for 0-based counter),or `(n..).zip` if the counter isn't 0-based.  (clippy::explicit_counter_loop)


Ph4ntomas:
Uh sorry what? Why are the tests failing on `just` setup? 

EDIT: I think GH Actions were impacted a bit by the current outage, but I can't restart the Action

Ottatop (approved):
Thanks!

PR #452, opened 2026-05-10
nixos/pinnacle: add protobuf files to system like home-manager

doronbehar:
Hello,

While trying to setup pinnacle as a NixOS service, using the service defined in `flake.nix`, I found this to be useful. Without it, a Lua configuration setup failed to launch with the error `could not find protobuf definitions directory`.


cassandracomar (approved):

Ottatop (approved):
Thanks!

PR #457, opened 2026-06-17
Update Cargo.toml to use HTTPS for pinnacle-api

thermosflasche:
Needed to change the URL to `https://`, otherwise `cargo build` on a freshly new `pinnacle config gen` won't properly fetch the repo


Ottatop (approved):
Thanks!

PR #458, opened 2026-06-21
feature: add PINNACLE_LOG_TIMESTAMPS env (fix #455)

### LINKED ISSUE #455: Add options to not log time stamps in stdout & stderr?
Hello,

I enjoy Pinnacle a lot on my embedded system. On my setup, I use `systemd-cat pinnacle` so that it'd be easier for me to view the logs of it via ssh. Both pinnacle & Systemd's Journal show me duplicate time stamps. I was wondering, would you be willing to accept a modification like this:

```diff
diff --git i/src/main.rs w/src/main.rs
index 738a796d..60433361 100644
--- i/src/main.rs
+++ w/src/main.rs
@@ -80,6 +80,7 @@ async fn main() -> anyhow::Result<()> {
     let stdout_layer = tracing_subscriber::fmt::layer()
         .compact()
         .with_writer(std::io::stdout)
+        .with_timer(env::var("PINNACLE_LOG_TIMESTAMPS").map(|v| v != "0" && v != "false").unwrap_or(true))
         .with_filter(stdout_env_filter);
 
     tracing_subscriber::registry()
```

doronbehar:
Other changes include:

- **pinnacle.service: add vim modeline**
- **running.md: fix small --session related typo**

### DIFF
--- a/resources/pinnacle.service
+++ b/resources/pinnacle.service
@@ -11,6 +11,8 @@ Before=xdg-desktop-autostart.target
 [Service]
 Slice=session.slice
 Type=notify
+# Spare duplicate (systemd & pinnacle) timestamps in log output, see src/main.rs
+Environment=PINNACLE_LOG_TIMESTAMPS=0
 ExecStart=/usr/bin/pinnacle --session
 
 # vim:ft=systemd
--- a/resources/pinnacle.service
+++ b/resources/pinnacle.service
@@ -12,3 +12,5 @@ Before=xdg-desktop-autostart.target
 Slice=session.slice
 Type=notify
 ExecStart=/usr/bin/pinnacle --session
+
+# vim:ft=systemd
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: GPL-3.0-or-later
 
 use std::{
-    env,
+    env, fmt,
     io::{BufRead, BufReader},
     sync::atomic::Ordering,
     time::Duration,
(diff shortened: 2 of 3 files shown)


doronbehar:
Sorry for exhausting CI a bit, now I tested it locally and it should at least build. I admit I haven't checked this in production. The test failure seems unrelated.

src/main.rs
```diff
@@ -78,6 +90,11 @@ async fn main() -> anyhow::Result<()> {
     let stdout_env_filter =
         env_filter.unwrap_or_else(|_| EnvFilter::new("warn,pinnacle=info,snowcap=info,sctk=error"));
     let stdout_layer = tracing_subscriber::fmt::layer()
+        .with_timer(OptionalTimer(
+            env::var("PINNACLE_LOG_TIMESTAMPS")
+                .map(|v| v != "0" && v != "false")
+                .unwrap_or(true),
+        ))
```
Ottatop, line 97: I believe using [`without_time`](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/struct.Layer.html#method.without_time) in an if statement is more appropriate here

resources/pinnacle.service
```diff
@@ -11,4 +11,8 @@ Before=xdg-desktop-autostart.target
 [Service]
 Slice=session.slice
 Type=notify
+# Spare duplicate (systemd & pinnacle) timestamps in log output, see src/main.rc
```
Ottatop:
```suggestion
# Spare duplicate (systemd & pinnacle) timestamps in log output, see src/main.rs
```

Ottatop (changes requested):

src/main.rs
doronbehar: > I believe using [`without_time`](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/struct.Layer.html#method.without_time) in an if statement is more appropriate here

@Ottatop I thought about it at first, but this turned out to be the most elegant way to do it, since the `.without_time()` method returns a layer of a different type, and it was too clumsy to use it. I am not fluent in Rust so if you have a concrete idea how to use it I'd be happy to hear.

Ottatop: Ah right, this is fine then

Ottatop (approved):
Thanks!

PR #459, opened 2026-06-21
Nix: filter src robustly with gitignore.nix

doronbehar:
I noticed this can be useful when playing with Nix files. The lack of sufficient `VERGEN_*` environment variables might tilt you towards not accepting it, but maybe I can think of different solutions - maybe without `gitignore.nix`.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -39,6 +39,26 @@
         "type": "github"
       }
     },
+    "gitignore": {
+      "inputs": {
+        "nixpkgs": [
+          "nixpkgs"
+        ]
+      },
+      "locked": {
+        "lastModified": 1762808025,
+        "narHash": "sha256-XmjITeZNMTQXGhhww6ed/Wacy2KzD6svioyCX7pkUu4=",
+        "owner": "hercules-ci",
+        "repo": "gitignore.nix",
+        "rev": "cb5e3fdca1de58ccbc3ef53de65bd372b48f567c",
+        "type": "github"
+      },
+      "original": {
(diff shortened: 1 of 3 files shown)


cassandracomar:
this makes it much harder to keep the package in-sync with what can be upstreamed to nixpkgs (PR pending). writing a real filter function for the source seems like a better idea.

doronbehar:
> this makes it much harder to keep the package in-sync with what can be upstreamed to nixpkgs (PR pending).

I had a very similar experience as here when [I helped `sile` write their Nix files](https://github.com/sile-typesetter/sile/issues?q=involves%3A%40me%20label%3Atooling), and I'm not sure what exactly makes it harder.

> writing a real filter function for the source seems like a better idea.

What isn't real with gitignore.nix? :)

cassandracomar:
we can't use gitignore.nix in nixpkgs

doronbehar:
> we can't use gitignore.nix in nixpkgs

But of course you won't need to. If you are talking about the experience of modifying the `nix/packages/default.nix` file make it work for `nixpkgs/pkgs/by-name/pi/pinnacle/package.nix`, even now you have to replace the `src` arguments in the current expression with a real `src = fetchFromGitHub`. This PR barely makes that experience less comfortable.

doronbehar:
What say you @cassandracomar ?

cassandracomar:
I still prefer writing a basic filter function and using the tools already in the nixpkgs lib (over pulling in a new dependency as a flake input).

PR #461, opened 2026-07-15
fix(nix): compat53 version compat

cassandracomar:
nixpkgs updated the version of compat53 to 0.15. this updates the package here [to match upstream](https://github.com/NixOS/nixpkgs/commit/9a2b9bf6becff18e47374d99cf4d674800a07a7f) so the package works with both pre-compat53 update and post-.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -8,11 +8,11 @@
         "rust-analyzer-src": "rust-analyzer-src"
       },
       "locked": {
-        "lastModified": 1772521798,
-        "narHash": "sha256-N69u6BZTfCyCsqyc1Qv/qvqkGkTCO6peGInKae51r/E=",
+        "lastModified": 1786005881,
+        "narHash": "sha256-Bj3VKIiwX1daO4vOIyAM507A65J0aaRSctc87kiT7H4=",
         "owner": "nix-community",
         "repo": "fenix",
-        "rev": "9ba6d89cd7cb4d2b94953d56bea1d46e08aa53bd",
+        "rev": "a7c3ef79859bf024cdc865a0ed8f14a4e3f31099",
         "type": "github"
       },
       "original": {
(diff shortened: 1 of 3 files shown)


doronbehar:
That's a bad AI PR if you ask me. Why not fix the version constraint for non-Nix users too - straight in https://github.com/pinnacle-comp/pinnacle/blob/main/api/lua/pinnacle-api-dev-1.rockspec ?

Ph4ntomas:
> That's a bad AI PR if you ask me.

There's no need for that. You don't know if AI was used or not. And even if that was the case, it's not helpful.

> Why not fix the version constraint for non-Nix users too - straight in https://github.com/pinnacle-comp/pinnacle/blob/main/api/lua/pinnacle-api-dev-1.rockspec ?

I agree with that part, but `>=` isn't the best without additional constraints, so maybe `~> 0.15` would be better here.

doronbehar:
> `>=` isn't the best without additional constraints, so maybe `~> 0.15` would be better here.

Maybe use:

```
compat53 >= 0.13, < 0.16,
```

cassandracomar:
> That's a bad AI PR if you ask me.

?? I linked the upstream commit this is based on.

> Why not fix the version constraint for non-Nix users too

I'm not positive whether that can be done without needing to cut a new release + bump the version. I figured this was a minimal change to unbreak the nix package. I don't even think this is broken if you're not using nix as you can still access the older version of compat53. nix just doesn't offer multiple versions of the same package by default, hence this patch step.

edit: I've also pinned back libdisplay-info to v0.3 as that's another breakage with this package (matching [nixpkgs PR](https://github.com/NixOS/nixpkgs/pull/549870)). clippy is breaking in CI -- not positive what's changed to prompt that yet.

Issue #464, opened 2026-08-01
Pinnacle fails to install on Debian 13


skorakora:
### Describe the bug

`just install` recipe fails on `install-lua-lib` because justfile replaces LUA_PATH and LUA_CPATH on the beggining of the just file.

### Reproduction steps

Run `just install` on Debian 13 (trixie)

### Version info

commit 12d0241ea6ef612d7e31596e4b2efb4c38565e83

### Distro

Debian 13

### GPU

not related

### Config language

Rust

PR #465, opened 2026-08-02
Fixed installer for debian 13

skorakora:
This pull request is related to issue #464

I still have not fully tested it because to test pinnacle I need to make it work on my debian first.
If someone that has more experience with pinnacle could test it I would be happy.
I have just tested that it builds and installs correctly and luarocks uses correct interpreter and lua header files (as luarocks is separate from lua and it will work with newer versions of lua interpretter).

What I did:
The problem was caused by problems with `justfile` altering with `LUA_PATH` and `LUA_CPATH`.
This caused luarocks to crash because it had missing depedencies when `LUA_PATH` was changed.
Another problem is that debian version of luarocks is not supporting header file location change by using `--lua-incdir` option.
More universal way was to change `LUA_INCDIR` path just for installing `pinnacle-api-dev-1.rockspec`

Enjoy pinnacle on debian :cat:


justfile
```diff
@@ -43,7 +25,7 @@ install-protos:
 install-lua-lib: gen-lua-pb-defs
     #!/usr/bin/env bash
     cd "{{rootdir}}/api/lua"
-    luarocks make --local --deps-mode "order" --lua-version "{{lua_version}}" pinnacle-api-dev-1.rockspec
+    luarocks make --local --deps-mode "order" --lua-version "{{lua_version}}" pinnacle-api-dev-1.rockspec "LUA_INCDIR=/usr/include/lua{{lua_version}}"
```
Ph4ntomas: This `LUA_INCDIR` path isn't a given.

On arch for example, latest version of lua is directly under `/usr/include` (which seems to be luarocks default too).
This will also break any manual lua installations under `/usr/local/include".

Could you check under `/etc/luarocks/` for your distro defaults? It kinda looks like either debian didn't package things properly or a bug in luarocks :/

In any case, if it´s fixed here, it would be better (imo) to conditionally override LUA_INCDIR.

Something like:
```
# this can be set by passing `with_lua_incdir=1` when invoking just
with_lua_incdir := ""
# provide a sane default, but let user override it.
lua_incdir_path := "/usr/include/lua" + lua_version

lua_incdir_path := if override_incdir != "" {
  "LUA_INCDIR=" + lua_incdir_path
} else { "" }
```

And then use that "{{lua_incdir_path}}" when luarocks is called.

skorakora: I did simple check what distro is used.
If it detects debian 13 it installs `pinnacle-api-dev-1.rockspec` as debian wants.
For other distros uses standard installation, but still I recomend testing it on other distros.
As comments suggest this can be used for any distro-specific installation, just follow the instructions in comments.

PS. I have added `TODO:` and `NOTE:` tags  as they are standard convention for developement called comment tags. You can follow this convention if you wish.
PSS. PLS resovle conversation if you think this will work thks :cat:

justfile
```diff
@@ -21,11 +21,42 @@ install-protos:
...
+    # TODO: Add more distro speciffic paths 
+    # TODO: Add support to custom locations
+    # TODO: Add testing for other distros
+
+    if [ "$ID" = "debian" ] && [ "$VERSION_ID" = "13" ]; then
+        echo "Debian 13 detected - installing it debian way..."
+        # Debian 13-specific installation
+
+        cd "{{rootdir}}/api/lua"
+        luarocks make --local --deps-mode "order" --lua-version "{{lua_version}}" pinnacle-api-dev-1.rockspec "LUA_INCDIR=/usr/include/lua{{lua_version}}"
```
Ph4ntomas, line 43: I don't think this is a good idea. As it stand, I can already tell that it's broken for every other Debian version, and likely for every debian-based distro (although I can't test it ATM.).

I would rather have a generic solution (even if that means asking the user to opt-in when invoking just) than having to maintain quirks for every distro.

justfile
```diff
@@ -21,11 +21,42 @@ install-protos:
...
 install-lua-lib: gen-lua-pb-defs
     #!/usr/bin/env bash
-    cd "{{rootdir}}/api/lua"
-    luarocks make --local --deps-mode "order" --lua-version "{{lua_version}}" pinnacle-api-dev-1.rockspec "LUA_INCDIR=/usr/include/lua{{lua_version}}"
+    . /etc/os-release
+
+    #simple condition that detects what distro it is
+    # TODO: Add more distro speciffic paths 
+    # TODO: Add support to custom locations
+    # TODO: Add testing for other distros
```
Ph4ntomas, line 36: I really don't think it's something to be done which is what TODOs tags are for.

The code essentially workaround a distro bug/tool misconfiguration, and workaround should not be considered code to replicate, but code to remove. 
If we're implementing workarounds, the comment should document why it was added, and under what condition it can be removed.

justfile
```diff
@@ -21,11 +21,42 @@ install-protos:
...
+
+    #else if [ "$ID" = <distro ID> ] && [ "$VERSION_ID" = <Distro version ID. ]; then
+    #    echo "<Distro name> detected - installing it <Distro name> way 
+    #    # Instalation specific to the specific distrubution...
+
+    else
+        echo "Debian 13 not detected - installing it normal way" # TODO: Change this later when other distros are implemented to avoid misinformation on logs
+        # Normal installation for luarocks that is distro independent and customizable
+
+        cd "{{rootdir}}/api/lua"
```
Ph4ntomas, line 56: If we keep the workaround for debian, can this `cd` be factored out?

justfile
```diff
@@ -21,11 +21,42 @@ install-protos:
...
+    # NOTE: This is an example of how to use this condition for other distributions.
+    # Feel free to add more. My work here is done, since I'm a Debian-only weirdo. //skorakora
+
+    #else if [ "$ID" = <distro ID> ] && [ "$VERSION_ID" = <Distro version ID. ]; then
+    #    echo "<Distro name> detected - installing it <Distro name> way 
+    #    # Instalation specific to the specific distrubution...
+
+    else
+        echo "Debian 13 not detected - installing it normal way" # TODO: Change this later when other distros are implemented to avoid misinformation on logs
+        # Normal installation for luarocks that is distro independent and customizable
```
Ph4ntomas, line 54: As a rule of thumb, when implementing a workaround, you shouldn't print when it's not being triggered. It looks like not having Debian 13 is unexpected.

justfile
```diff
@@ -21,11 +21,42 @@ install-protos:
...
+
+        cd "{{rootdir}}/api/lua"
+        luarocks make --local --deps-mode "order" --lua-version "{{lua_version}}" pinnacle-api-dev-1.rockspec "LUA_INCDIR=/usr/include/lua{{lua_version}}"
+
+    # NOTE: This is an example of how to use this condition for other distributions.
+    # Feel free to add more. My work here is done, since I'm a Debian-only weirdo. //skorakora
+
+    #else if [ "$ID" = <distro ID> ] && [ "$VERSION_ID" = <Distro version ID. ]; then
+    #    echo "<Distro name> detected - installing it <Distro name> way 
+    #    # Instalation specific to the specific distrubution...
```
Ph4ntomas, line 50: Same as for the `TODO` above, I don't think this is warranted. 
Beside that there's not need to clutter the file with comments essentially saying "you can duplicate the code from the workaround by duplicating the code from the workaround, and tweaking it"

justfile
```diff
@@ -21,11 +21,42 @@ install-protos:
     mkdir -p "{{xdg_data_dir}}"
     cp -r "{{rootdir}}/api/protobuf" "${proto_dir}"
 
+
+
+
+
```
Ph4ntomas, line 27: Could you remove those?

Ph4ntomas (changes requested):

